PR #52, opened 2017-04-27
Bump rancher compose executor

joshwget:


deniseschannon:
https://github.com/rancher/rancher/issues/8598

PR #63, opened 2017-06-19
Fix wait for up

StrongMonkey:
Interesting Point:
When we launch the stack in the first time, the `wait` won't actually work because at the time when we call `wait` the stack is created but none of the service is launched so the stack is actually healthy. But we still wait at [service creation level](https://github.com/rancher/rancher-compose-executor/blame/master/rancher/service.go#L156-L159).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,5 @@
 /dist
 *.swp
 /.trash-cache
+/.idea
+/cli
--- a/cmd/up.go
+++ b/cmd/up.go
@@ -1,6 +1,8 @@
 package cmd
 
 import (
+	"github.com/pkg/errors"
+	"github.com/rancher/go-rancher/v2"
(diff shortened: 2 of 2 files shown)


StrongMonkey:
https://github.com/rancher/rancher/issues/8492

StrongMonkey:
There is not test because we don't have an integration test framework for our CLI project.

.gitignore
```diff
@@ -4,3 +4,5 @@
 /dist
 *.swp
 /.trash-cache
+/.idea
```
cjellick: Dont bput .idea in the project's .gitignore. put it in your global git ignore file that can live at the root of your home directory.

StrongMonkey: 👌

deniseschannon:
https://github.com/rancher/cli/pull/73

PR #65, opened 2017-06-22
add error msg for invalid -e option

carolynhu:
https://github.com/rancher/rancher/issues/9132

### DIFF
--- a/cmd/run.go
+++ b/cmd/run.go
@@ -5,6 +5,7 @@ import (
 
 	"github.com/rancher/go-rancher/v2"
 
+	"fmt"
 	"github.com/urfave/cli"
 	"os"
 )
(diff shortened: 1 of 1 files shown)


cmd/run.go
```diff
@@ -294,8 +295,14 @@ func serviceRun(ctx *cli.Context) error {
 	for _, env := range ctx.StringSlice("env") {
 		parts := strings.SplitN(env, "=", 2)
 		value := ""
+
 		if len(parts) > 1 {
 			value = parts[1]
+
+			if len(parts[0]) == 0 {
```
joshwget: Can you switch to `parts[0] != ""`? A bit more idiomatic.

cmd/run.go
```diff
@@ -294,8 +295,14 @@ func serviceRun(ctx *cli.Context) error {
...
+				errMsg := fmt.Sprintf("invalid argument \"%s\" for e: invalid environment variable: %s\nSee 'rancher run --help'.", env, env)
```
joshwget: This is a bit verbose. Maybe we can just use the second half of the error?

carolynhu: I follow docker's error message on that, if we run: docker run -e "=bar" ubuntu, the msg will be:
invalid argument "=bar" for e: invalid environment variable: =bar
See 'docker run --help'.

cmd/run.go
carolynhu: sure

joshwget:
LGTM

PR #66, opened 2017-06-28
Let users query all environments

raphink:
This is probably not the greatest implementation, but at least it'll get the discussion going.

The goal is here to be able to query all environments, to e.g. find in which environment an host is located. With this patch, you can use:

```shell
$ rancher --env ALL host   --format '{{.ID}} {{.Host.Hostname}} {{.Host.AccountId}}'
```

to retrieve that info. It also allows to set `environment` to `"ALL"` in the json configuration file for an account API key.

The choice for "ALL" instead of an empty string is because the current behavior with an empty environment is to list all existing environments.

### DIFF
--- a/cmd/common.go
+++ b/cmd/common.go
@@ -111,8 +111,12 @@ func GetEnvironment(def string, c *client.RancherClient) (*client.Project, error
 		for _, p := range resp.Data {
 			names = append(names, fmt.Sprintf("%s(%s)", p.Name, p.Id))
 		}
+		names = append(names, "ALL")
 
 		idx := selectFromList("Environments:", names)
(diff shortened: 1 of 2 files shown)


raphink:
@deniseschannon any opinion on this?

deniseschannon:
@raphink I need to look into this, but probably won't get to it until next week. But it's on my radar!

raphink:
Thanks

cjellick:
Hm. Seems to me like this won't work. If you don't provide an environment, you don't see everything across all environments in the API, you see nothing.

cjellick:
Very stale. Closing. Thanks! Please open a new PR if there is still a relevant change.

PR #71, opened 2017-07-27
WIP: update vendor file

amy:
https://github.com/rancher/rancher/issues/9305


deniseschannon:
I've set this PR as WIP as usually I like to wait until a rancher-compose-executor release is out so that the vendor.conf can be updated to a specific version and then update the vendor files at that time.

PR #73, opened 2017-08-01
Respect wait-state flag

amy:
https://github.com/rancher/rancher/issues/8492

### DIFF
--- a/cmd/up.go
+++ b/cmd/up.go
@@ -29,7 +29,19 @@ func UpCommand() cli.Command {
 			Usage: "Specify an alternate project name (default: directory name)",
 		},
 	}...)
+
+	cmd.Action = app.WithProject(factory, ProjectUp)
+
 	return cmd
+}
+
+func ProjectUp(p *project.Project, c *cli.Context) error {
+	w, err := NewWaiter(c)
+	if err != nil {
+		return err
+	}
+
+	return app.ProjectUpAndWait(p, w, c)
 }
 
 type projectFactory struct {
--- a/cmd/wait.go
+++ b/cmd/wait.go
@@ -8,6 +8,7 @@ import (
 	"github.com/Sirupsen/logrus"
 	"github.com/rancher/cli/monitor"
 	"github.com/rancher/go-rancher/v2"
+	"github.com/rancher/rancher-compose-executor/project/options"
 	"github.com/urfave/cli"
 )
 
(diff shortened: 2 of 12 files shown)


vendor/github.com/rancher/rancher-compose-executor/rancher/wait.go
```diff
@@ -22,9 +22,21 @@ func (r *RancherService) WaitFor(resource *client.Resource, output interface{},
 }
 
 func (r *RancherService) Wait(service *client.Service) error {
-	return r.WaitFor(&service.Resource, service, func() string {
+
+	err := r.WaitFor(&service.Resource, service, func() string {
 		return service.Transitioning
```
StrongMonkey: In `r.WaitFor` method it is already waiting for the service resource to become healthy(which waits its transition state). Why do we need to activate the service then apply the wait function again?

cmd/up.go
```diff
@@ -46,6 +48,16 @@ func (p *projectFactory) Create(c *cli.Context) (*project.Project, error) {
 		return nil, err
 	}
 
+	rc, err := GetClient(c)
+	if err != nil {
+		return nil, err
```
cjellick: Use `errors.Wrap`.
there are examples of its use in other places

cmd/wait.go
```diff
@@ -154,7 +160,7 @@ func (w *Waiter) Wait() error {
 		}
 
 		ok, err := w.done(r.Type, r.Id)
-		if err != nil {
+		if err != nil && w.state == "" {
```
amy: Please provide input for this line. Without `w.state == ""`, if the state is set, and the stack reaches an error while transitioning, it will still return from the cli. With this check, it'll continue waiting until the `--wait-flag` state agreed upon is reached.

amy:
Please use the following scenarios for manual testing: https://gist.github.com/amy/f34edd7e599363ec8789b3971e24c4e6

cmd/up.go
```diff
@@ -60,5 +72,10 @@ func (p *projectFactory) Create(c *cli.Context) (*project.Project, error) {
 	}
 
 	factory := &app.RancherProjectFactory{}
-	return factory.Create(c)
+	proj, err := factory.Create(c)
```
cjellick: if factory.Create(c) returns the same this as this method is returning, can't you just do
```
return factory.Create(c)
```

cmd/up.go
```diff
@@ -29,9 +29,21 @@ func UpCommand() cli.Command {
 			Usage: "Specify an alternate project name (default: directory name)",
 		},
 	}...)
+
+	cmd.Action = app.WithProject(factory, ProjectUp)
```
cjellick: I don't really know this codebase well, but having to set the Action here kind of scares me. Seems like this bit of code shouldnt need changed. What's the reason for changing it?

amy: Darren suggested this. So the idea is to override the action in order to inject the waiter in the cli part of the code (vs. in rancher-compose-executor code). So ultimately, we need to use the new `ProjectUp` function to inject the waiter.

cmd/up.go
amy: Good point. Changed that.

cjellick:
LGTM

PR #74, opened 2017-08-07
throw error if we try to parse an answers file but don't find any

wjimenez5271:


wjimenez5271:
https://github.com/rancher/rancher/issues/9665

cjellick:
Offline discussion determine that the proper fix is going to be to regex each line.

wjimenez5271:
@cjellick see above refactor using regex

wjimenez5271:
this PR doesn't apply any more since the refactor

PR #76, opened 2017-08-10
Update vendor for StopTimeout

carolynhu:
https://github.com/rancher/rancher/issues/7715


deniseschannon:
please break this up into 2 commits.

first commit for the updates in vendor.conf labeled with what versions you're updating to for rancher-compose-executor and that you're updating go-rancher

second commit would be the vendor updates

cjellick:
LGTM

PR #85, opened 2017-08-29
rancher-compose-executor v0.14.18 & catalog install & up envfile multiline

amy:
https://github.com/rancher/rancher/issues/9305#issuecomment-325463752


cjellick:
LGTM

PR #97, opened 2017-10-11
fix windows build

StrongMonkey:


StrongMonkey:
https://github.com/rancher/rancher/issues/10026

PR #102, opened 2017-10-23
Use  /project API to find environment instead of /account API

loganhz:
If a non-admin user calls the /account API, it will not return any projects. So rancher cli can not find the environment by environment name. If we use the /project API, it will return all the environments the user has. But it will only return parts of environments for an admin user if the request doesn't have the parameter all=true. So we have to add this query parameter in this case.

https://github.com/rancher/rancher/issues/9988


StrongMonkey:
LGTM

PR #105, opened 2017-12-11
Improve shebang

borisbabic:
It is generally better to use `#!/usr/bin/env bash`  than `#!/bin/bash`. `/bin/bash` isn't available everywhere.


cjellick:
Very stale. Closing. Thanks! Please open a new PR if there is still a relevant change.

PR #108, opened 2018-01-11
Feat: Deploy catalog packages that uses go templating at docker compose

rawmind0:
Catalog packages that uses go templating, docker-compose.yml.tpl, are not deployed correctly from the rancher cli. This PR solves it.


StrongMonkey:
LGTM

PR #110, opened 2018-01-18
Use Named Pipeline for Windows instead of Unix Socket

orangedeng:
Related issue rancher/rancher#7262
The unix proto is not supported in Windows.
We need to use named pipeline in Windows instead.

The vendor.conf needs to be modified because PR https://github.com/rancher/rancher-docker-api-proxy/pull/6 is not merged yet


orangedeng:
@deniseschannon this pr needs to be reviewed

ibuildthecloud:
LGTM

orangedeng:
@deniseschannon updated.

PR #115, opened 2018-02-13
CLI for 2.0 preview

dramich:
This gets the cli working with v2.0.0-alpha16 with a small set of commands.


dramich:

PR #116, opened 2018-02-15
Update to match UI 

dramich:
* Use the new nodeCommand
* Create blank clusters to match UI flow


dramich:

PR #117, opened 2018-02-15
Add delete methods to cluster and project

dramich:


dramich:

PR #132, opened 2018-03-15
Update log to use log action

dramich:
Problem:
Restricted users were not able to view logs due to the use of
docker.sock

Solution:
Move log calls to use the log action for the container

Issue: https://github.com/rancher/rancher/issues/8020


cjellick:
LGTM

PR #134, opened 2018-03-20
Update exec to use websocket

dramich:
Problem:
    There is not parity from the UI to the CLI for what type of user can use
    exec

Solution:
    Update exec to the use a websocket provided by the actions living on the
    container. This allows the api server to determine who has access to a
    container


cjellick:
LGTM

deniseschannon:
https://github.com/rancher/rancher/issues/8020

PR #135, opened 2018-03-21
Update README.md added 'rancher config' cmd

thenets:
Added the "rancher config" as the correctly way to setup Rancher CLI.


dramich:
Master is for rancher 2.0, which has a different set of commands. `rancher login` is the correct command for master.

thenets:
Oh, sorry. My mistake. I'll pay more attention to it next time.

PR #142, opened 2018-03-28
Golang 1.9.4

m4ce:
I had to upgrade to golang 1.9.4 to specifically SSL CA certs verification when using a custom CA authority.

FATA[0007] Get https://rancher/v2-beta/projects/1a5: x509: certificate signed by unknown authority

I am using OSX and this was fixed in golang 1.9.x. For more information, see docker/for-mac#2201.

Unfortunately, previous versions of golang wouldn't consider CA certs stored in the System keychain. They would only look into the System Roots one, which is not writable.

### DIFF
--- a/Dockerfile.dapper
+++ b/Dockerfile.dapper
@@ -1,4 +1,4 @@
-FROM golang:1.7.3
+FROM golang:1.9.4
 RUN apt-get update && \
     apt-get install -y xz-utils zip rsync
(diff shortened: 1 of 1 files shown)


deniseschannon:
Closed in favor of bumping to go 1.13.9 in this PR https://github.com/rancher/cli/pull/271

PR #143, opened 2018-03-28
Update kubectl to use cluster kubeconfig

dramich:
Problem:
kubectl is using the users kubeconfig on the system and not the one
specific to the cluster

Solution:
kubectl now pulls the clusters kubeconfig and uses that to run kubectl
commands


deniseschannon:
https://github.com/rancher/rancher/issues/12316

PR #145, opened 2018-03-29
Add description flag to project and namespace

dramich:
Problem:
When creating a project and namespace no description can be set

Solution:
Add a flag for optionally adding a description

Issue: https://github.com/rancher/rancher/issues/11461


cjellick:
LGTM

PR #146, opened 2018-03-30
Fix version list

dramich:
Problem:
When running `cluster create --k8s-version list` the cluster is still
being created

Solution:
After printing the versions exit

Issue: https://github.com/rancher/rancher/issues/12383


cjellick:
LGTM when green

PR #154, opened 2018-04-16
Update project create and ps to use lookup by name

dramich:
Problem:
Specifying a cluster on project create and a project using ps required
an ID

Solution:
Use the lookup by name
Update the description of the ps command

Issue: https://github.com/rancher/rancher/issues/11469


cjellick:
LGTM

PR #159, opened 2018-04-23
Fix container exec

dramich:
Problem:
Unable to run exec using the container name/ID

Solution:
Update logic to allow specifying the container name/ID

Issue: https://github.com/rancher/rancher/issues/12918


cjellick:
LGTM assuming green build

PR #173, opened 2018-05-17
Add ability to install charts locally

StrongMonkey:
need https://github.com/rancher/types/pull/437

### DIFF
--- a/cmd/app.go
+++ b/cmd/app.go
@@ -2,10 +2,13 @@ package cmd
 
 import (
 	"bufio"
+	"encoding/base64"
 	"encoding/json"
 	"fmt"
+	"io/ioutil"
 	"net/url"
 	"os"
+	"path/filepath"
 	"sort"
 	"strings"
 	"time"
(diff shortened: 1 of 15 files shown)


StrongMonkey:
https://github.com/rancher/rancher/issues/13518

cmd/app.go
```diff
@@ -17,6 +17,10 @@ import (
 	managementClient "github.com/rancher/types/client/management/v3"
 	projectClient "github.com/rancher/types/client/project/v3"
 	"github.com/urfave/cli"
+	"path/filepath"
```
dramich: Sort these, go imports should be in the top group

cmd/app.go
```diff
@@ -843,3 +898,20 @@ func checkShowSubquestionIf(q managementClient.Question, answers map[string]stri
 	}
 	return false
 }
+
+type ChartVersion struct {
```
dramich: Keep types together at the top of the file

PR #174, opened 2018-05-17
Wait for namespace to be ready

dramich:
Problem:
An app creates a namespace to deploy into and depending on the system it
might not be ready when attempting to install the app

Solution:
Poll the namespace to ensure it's active before proceeding with
installing the app

Issue: https://github.com/rancher/rancher/issues/13567

### DIFF
--- a/cmd/app.go
+++ b/cmd/app.go
@@ -16,6 +16,7 @@ import (
 	clusterClient "github.com/rancher/types/client/cluster/v3"
 	managementClient "github.com/rancher/types/client/management/v3"
 	projectClient "github.com/rancher/types/client/project/v3"
+	"github.com/sirupsen/logrus"
 	"github.com/urfave/cli"
 )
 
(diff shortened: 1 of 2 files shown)


cmd/app.go
```diff
@@ -687,10 +688,29 @@ func createNamespace(c *cliclient.MasterClient, n string) error {
...
+			ns, err = c.ClusterClient.Namespace.ByID(ns.ID)
+			if err != nil {
+				return err
+			}
+
+			if ns.State == "active" {
+				break
+			}
+
+			time.Sleep(1 * time.Second)
```
cjellick: My opinion: sleeping a second is too long and waiting an entire minute before erroring out is too long.  Half a second and 30 total would be slightly better?

Other than that, looks good

dramich: Updated to half a second and timeout at 30 seconds

cjellick:
LGTM

PR #178, opened 2018-06-11
use managementClient instead of masterClient

StrongMonkey:


StrongMonkey:
https://github.com/rancher/rancher/issues/13353

StrongMonkey:
https://github.com/rancher/rancher/issues/13361

PR #179, opened 2018-06-26
Export cluster

dramich:
Problem:
Cluster is not able to be exported

Solution:
Add sub command `rancher cluster export [name/ID]` to output a cluster yaml. 
This also unhides `up`

Issue: https://github.com/rancher/rancher/issues/14160


cjellick:
LGTM

PR #185, opened 2018-07-18
Allow ssh via external ip

jeremyweber-np:
Adds a -e flag to allow connecting to the external ip address of the node.  Also changes the username flag from -l to -u, which seems more appropriate from a rancher standpoint, but not as much from an ssh standpoint.  ???

### DIFF
--- a/cmd/ssh.go
+++ b/cmd/ssh.go
@@ -20,15 +20,21 @@ import (
 )
 
 const sshDescription = `
-For any nodes created through Rancher using docker-machine, 
+For any nodes created through Rancher using docker-machine,
 you can SSH into the node. This is not supported for any custom nodes.
 
 Examples:
 	# SSH into a node by ID/name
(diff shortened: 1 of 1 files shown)


dhrp:
@dramich any feedback or comments?

cmd/ssh.go
```diff
@@ -20,59 +20,68 @@ import (
...
 Examples:
 	# SSH into a node by ID/name
 	$ rancher ssh nodeFoo
 
+	# SSH into a node by ID/name using the external IP address
+	$ rancher ssh -e nodeFoo
+
 	# SSH into a node by name but specify the username to use
-	$ rancher ssh -l user1 nodeFoo
+	$ rancher ssh -u user1 nodeFoo
```
dramich: This flag should stay as is, it matches ssh usage.

dhrp: Ok, I've reverted that and made a new PR

PR #187, opened 2018-07-30
Make context id configurable at login

superseb:
https://github.com/rancher/rancher/issues/14448

### DIFF
--- a/cmd/login.go
+++ b/cmd/login.go
@@ -41,9 +41,9 @@ func LoginCommand() cli.Command {
 		Action:    loginSetup,
 		ArgsUsage: "[SERVERURL]",
 		Flags: []cli.Flag{
-			cli.BoolFlag{
+			cli.StringFlag{
 				Name:  "context",
-				Usage: "Get the currently set context",
+				Usage: "Set the context during login",
 			},
 			cli.StringFlag{
 				Name:  "token,t",
@@ -62,14 +62,6 @@ func LoginCommand() cli.Command {
 }
 
 func loginSetup(ctx *cli.Context) error {
-	if ctx.Bool("context") {
-		err := loginContext(ctx)
-		if nil != err {
-			return err
-		}
-		return nil
-	}
-
 	if ctx.NArg() == 0 {
 		return cli.ShowCommandHelp(ctx, "login")
 	}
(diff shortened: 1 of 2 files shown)


cmd/login.go
```diff
@@ -41,9 +41,9 @@ func LoginCommand() cli.Command {
 		Action:    loginSetup,
 		ArgsUsage: "[SERVERURL]",
 		Flags: []cli.Flag{
-			cli.BoolFlag{
+			cli.StringFlag{
 				Name:  "context",
-				Usage: "Get the currently set context",
+				Usage: "Set the context after login",
```
dramich: "Set the context during login"

cmd/login.go
```diff
@@ -135,7 +127,9 @@ func loginSetup(ctx *cli.Context) error {
 		}
 	}
 
-	proj, err := getProjectContext(ctx, c)
+	project := ctx.String("project")
```
dramich: There is no flag called `project` here so it's just an empty string.

cmd/login.go
```diff
@@ -150,7 +144,7 @@ func loginSetup(ctx *cli.Context) error {
 	return nil
 }
 
-func getProjectContext(ctx *cli.Context, c *cliclient.MasterClient) (string, error) {
+func getProjectContext(ctx *cli.Context, c *cliclient.MasterClient, project string) (string, error) {
```
dramich: `project string` is never used so should be removed

cmd/login.go
```diff
@@ -161,6 +155,22 @@ func getProjectContext(ctx *cli.Context, c *cliclient.MasterClient) (string, err
 		return "", nil
 	}
 
+	// If project is given
+	if ctx.String("context") != "" {
```
dramich: Move this to the top of the function, if user gives `context` let's assume they know there is an existing project to use and not do the list.

cmd/login.go
```diff
@@ -161,6 +155,22 @@ func getProjectContext(ctx *cli.Context, c *cliclient.MasterClient) (string, err
...
+		context := ctx.String("context")
+		// Check if given context is in valid format
+		_, _, err := ParseClusterAndProjectID(context)
+		if err != nil {
+			return "", fmt.Errorf("Unable to parse context (%s). Please provide context as c-xxxxx:p-xxxxx or c-xxxxx:project-xxxxx", context)
+		}
+		// Check if context exists
+		_, err = Lookup(c, context, "project")
+		if err != nil {
+			return "", fmt.Errorf("Unable to find context (%s). Make sure the context exists and you have permissions to use it", context)
```
dramich: Include the error returned from `Lookup`

dramich:
LGTM

PR #189, opened 2018-09-06
Add quiet option to add-node and import,deprecate use of --management

superseb:
https://github.com/rancher/rancher/issues/15441
https://github.com/rancher/rancher/issues/15442


dramich:
LGTM

PR #191, opened 2018-09-11
Add pagination to resourceLs functions

superseb:
https://github.com/rancher/rancher/issues/15453
https://github.com/rancher/rancher/issues/9663

### DIFF
--- a/cmd/catalog.go
+++ b/cmd/catalog.go
@@ -450,7 +450,20 @@ func (c *CatalogOperator) forEachTemplate(f func(item *catalog.Template) error) 
 		return err
 	}
 
-	for _, item := range collection.Data {
+	collectiondata := collection.Data
+
+	for {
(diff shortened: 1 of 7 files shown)


cmd/catalog.go
```diff
@@ -450,7 +450,20 @@ func (c *CatalogOperator) forEachTemplate(f func(item *catalog.Template) error)
 		return err
 	}
 
-	for _, item := range collection.Data {
+	collectiondata := collection.Data
```
alena1108: this code seems to repeat in several places. Any way we can extract it to some helper method, and call from there?

alena1108: Looked at the code more, seems like implementing it in generic manner using interfaces would be even more cumbersome.

alena1108:
LGTM

PR #192, opened 2018-09-11
Update namespace move to use action

dramich:
Problem:
Namespace move is now an action on a namespace

Solution:
Update cli to use the namespace action if it exists, otherwise use update


cmd/namespace.go
```diff
@@ -171,7 +169,7 @@ func namespaceDelete(ctx *cli.Context) error {
 }
 
 func namespaceAssociate(ctx *cli.Context) error {
```
StrongMonkey: also change function name?

dramich: Good idea, updated.

cmd/namespace.go
```diff
@@ -190,8 +188,25 @@ func namespaceAssociate(ctx *cli.Context) error {
...
+	if _, ok := namespace.Actions["move"]; ok {
+		move := &clusterClient.NamespaceMove{
+			ProjectID: proj.ID,
+		}
+		return c.ClusterClient.Namespace.ActionMove(namespace, move)
+	}
+
 	update := make(map[string]string)
-	update["projectId"] = ctx.Args().Get(1)
+	update["projectId"] = proj.ID
```
StrongMonkey: Don't have to update projectID. Move action should update it. And also projectID is non-updatable.

dramich: This happens on versions before your change.

StrongMonkey:
LGTM

cmd/namespace.go
cjellick: @dramich there is a scenario where the new version of rancher is used AND the move action is not present.
THat is actually the primary reason for making it an action: so that the UI can pre-emptively tell if the namesapce can be moved.

So, I like that you added the backwards compatibility, but the logic might have to get more complicated.

I don;t know the solutoin, but here's my rough idea on how it should behave:
1. if move action is available, call it.
2. for backwards compat, if it is not available, you can try update, but update, I believe, will silently drop the update and return. So, when update returns, if the project ID remains the same, tell the user that the ns may not be moved

dramich: Fixed by checking the annotations

PR #193, opened 2018-09-13
Improve error when there are no projects found

superseb:


cjellick:
LGTM but needs an issue @superseb

PR #202, opened 2018-10-04
Enhance scripting with cli by adding wait

dramich:
Problem:
CLI asks to verify certs on login
No way to wait for a resource to be available

Solution:
Add a skip-verify flag to login
Add wait command to wait for a cluster to be in active state

Issue: https://github.com/rancher/rancher/issues/15692

### DIFF
--- a/cmd/wait.go
+++ b/cmd/wait.go
@@ -1,0 +1,89 @@
+package cmd
+
+import (
+	"fmt"
+	"strings"
+	"time"
+
+	ntypes "github.com/rancher/norman/types"
+	"github.com/sirupsen/logrus"
+	"github.com/urfave/cli"
+)
+
+var (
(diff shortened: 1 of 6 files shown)


StrongMonkey:
LGTM

cmd/wait.go
```diff
@@ -0,0 +1,89 @@
...
+		return cli.ShowCommandHelp(ctx, "wait")
+	}
+
+	c, err := GetClient(ctx)
+	if err != nil {
+		return err
+	}
+
+	resource, err := Lookup(c, ctx.Args().First(), waitTypes...)
+	if nil != err {
```
StrongMonkey: Nit: err != nil might have a better look

cjellick:
LGTM

PR #203, opened 2018-10-18
Update logs command

dramich:
This fix along with the fix to [agent](https://github.com/rancher/agent/pull/284) will get `tail/lines` working but additional work in cattle-server will be needed to get the rest of the options working. 

https://github.com/rancher/rancher/issues/12924

### DIFF
--- a/cmd/logs.go
+++ b/cmd/logs.go
@@ -18,8 +18,10 @@ import (
 )
 
 type logOptions struct {
-	Follow bool `json:"follow"`
-	Lines  int  `json:"lines"`
+	Follow     bool   `json:"follow"`
+	Lines      int    `json:"lines"`
+	Timestamps bool   `json:"timestamps"`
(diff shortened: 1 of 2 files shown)


cmd/logs.go
```diff
@@ -241,7 +245,13 @@ func doLog(wsURL string) error {
 		if nil != err {
 			return err
 		}
-		fmt.Println(string(buf))
+		// The websocket frame has "0* " to start the log, drop that to match UI
+		log := strings.SplitN(string(buf), " ", 2)
+		if len(log) == 2 {
+			fmt.Print(log[1])
```
StrongMonkey: should we use println?

StrongMonkey: Can we make sure the log always ends with `/n`?

StrongMonkey:
LGTM

PR #204, opened 2018-10-23
handle forbidden error when checking new namespace

gitlawr:
Address Issue https://github.com/rancher/rancher/issues/16132

### DIFF
--- a/cmd/app.go
+++ b/cmd/app.go
@@ -6,6 +6,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"io/ioutil"
+	"net/http"
 	"net/url"
 	"os"
 	"path/filepath"
@@ -16,6 +17,7 @@ import (
 	gover "github.com/hashicorp/go-version"
 	"github.com/pkg/errors"
 	"github.com/rancher/cli/cliclient"
+	"github.com/rancher/norman/clientbase"
 	clusterClient "github.com/rancher/types/client/cluster/v3"
 	managementClient "github.com/rancher/types/client/management/v3"
 	projectClient "github.com/rancher/types/client/project/v3"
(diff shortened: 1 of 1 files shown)


cmd/app.go
```diff
@@ -5,7 +5,9 @@ import (
 	"encoding/base64"
 	"encoding/json"
 	"fmt"
+	"github.com/rancher/norman/clientbase"
```
orangedeng: This might have a go fmt problem

orangedeng:
LGTM

cmd/app.go
```diff
@@ -787,14 +789,18 @@ func createNamespace(c *cliclient.MasterClient, n string) error {
...
 		startTime := time.Now()
 		for {
 			logrus.Debug(fmt.Sprintf("Namespace create wait - Name: %s, State: %s, Transitioning: %s", ns.Name, ns.State, ns.Transitioning))
 			if time.Since(startTime)/time.Second > 30 {
 				return fmt.Errorf("timed out waiting for new namespace %s", ns.Name)
 			}
-			ns, err = c.ClusterClient.Namespace.ByID(ns.ID)
-			if err != nil {
+			ns, err = c.ClusterClient.Namespace.ByID(nsID)
+			if e, ok := err.(*clientbase.APIError); ok && e.StatusCode == http.StatusForbidden {
```
dramich: This check should be inside the `if err != nil {` if we don't have an error we don't need to type assert the error.
```
			ns, err = c.ClusterClient.Namespace.ByID(nsID)
			if err != nil {
				if e, ok := err.(*clientbase.APIError); ok && e.StatusCode == http.StatusForbidden {
					continue
				}
				return err
			}
```

gitlawr: make sense

dramich:
LGTM

PR #205, opened 2018-10-23
add no-prompt option for app install

gitlawr:
The option will be helpful for scripting using cli.
Address issue https://github.com/rancher/rancher/issues/16177

### DIFF
--- a/cmd/app.go
+++ b/cmd/app.go
@@ -145,6 +145,10 @@ func AppCommand() cli.Command {
 						Name:  "version",
 						Usage: "Version of the template to use",
 					},
+					cli.BoolFlag{
+						Name:  "no-prompt",
+						Usage: "Suppress asking questions and use the default values when required answers are not provided",
+					},
 				},
 			},
 			cli.Command{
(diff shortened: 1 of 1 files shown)


cmd/app.go
```diff
@@ -145,6 +145,10 @@ func AppCommand() cli.Command {
 						Name:  "version",
 						Usage: "Version of the template to use",
 					},
+					cli.BoolFlag{
+						Name:  "no-prompt",
+						Usage: "Suppress asking questions and use the default values when required answers are not provided",
```
alena1108: what is the default value?

cmd/app.go
```diff
@@ -561,8 +565,9 @@ func templateInstall(ctx *cli.Context) error {
 			return err
 		}
 
+		interactive := !ctx.Bool("no-prompt")
 		answers := make(map[string]string)
-		err = processAnswers(ctx, c, templateVersion, answers, true)
+		err = processAnswers(ctx, c, templateVersion, answers, interactive)
```
alena1108: if no-prompt is set to false by default, the old behavior of passing true to processAnswers by default won't be preserved.

cmd/app.go
gitlawr: False for bool value

cmd/app.go
gitlawr: There's a negative above, interactive == !no-prompt

alena1108: ah, my bad

cmd/app.go
alena1108: shall we list the default value in help? there is a cli flag for it, called default value or something

PR #207, opened 2018-10-25
Allow equal signs to appear in the value part of an answer

brandonbodnar-wk:
If an answer had an equals sign within the value portion of the answer, then the answer would be silently ignored and not included in the request to the rancher server.

For example,

```console
$ rancher app install --set my.foo="a-string-with-a-=-sign" . my-app
```

Would attempt to install the chart as my-app, but the request payload would not include my.foo.

Change respects only the first equal sign within the answer and then takes all that follows as the value.

### DIFF
--- a/cmd/app.go
+++ b/cmd/app.go
@@ -832,7 +832,7 @@ func processAnswers(
 	}
 
 	for _, answer := range ctx.StringSlice("set") {
-		parts := strings.Split(answer, "=")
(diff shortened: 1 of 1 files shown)


JasonvanBrackel:
@deniseschannon Can we get this PR'd and pushed by an engineer?

StrongMonkey:
LGTM

alena1108:
https://github.com/rancher/rancher/issues/16405

PR #209, opened 2018-11-07
fix app upgrade show-versions option

gitlawr:
Address issue: https://github.com/rancher/rancher/issues/16478

### DIFF
--- a/cmd/app.go
+++ b/cmd/app.go
@@ -290,7 +290,7 @@ func appUpgrade(ctx *cli.Context) error {
 		return err
 	}
 
-	if ctx.Bool("versions") {
+	if ctx.Bool("show-versions") {
 		return outputVersions(ctx, c)
 	}
 


orangedeng:
LGTM

StrongMonkey:
LGTM

cmd/app.go
```diff
@@ -290,7 +290,7 @@ func appUpgrade(ctx *cli.Context) error {
 		return err
 	}
 
-	if ctx.Bool("versions") {
+	if ctx.Bool("show-versions") {
```
alena1108: @dramich just wanted to confirm that the intended param name as we have it `versions` in one place, and `show-versions` in another. As it never worked anyway, we have flexibility to decide on the name now

dramich: @alena1108 I think leaving the flag as `show-versions` is fine, but I am not stuck on it. If you want to change it to `versions` that's fine as well.

PR #214, opened 2019-01-21
Add multi-cluster app support

gitlawr:
Address issue: https://github.com/rancher/rancher/issues/17147


gitlawr:
@orangedeng Please help review this PR, thanks

cmd/multiclusterapp.go
```diff
@@ -0,0 +1,668 @@
...
+		return err
+	}
+
+	for _, arg := range ctx.Args() {
+		resource, err := Lookup(c, arg, managementClient.MultiClusterAppType)
+		if err != nil {
+			return err
+		}
+
+		app, err := c.ManagementClient.MultiClusterApp.ByID(resource.ID)
```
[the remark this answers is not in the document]
orangedeng: I see that the other delete commend will do it in this way, but the `.Delete()` function only requires `type.Resource` in the object. Maybe we only need to wrap the `resource` returned by `Lookup` function instead of using `ByID` to get it.

cmd/multiclusterapp.go
```diff
@@ -0,0 +1,703 @@
...
+						Name:  "set",
+						Usage: "Set answers for the template, can be used multiple times. Example: --set foo=bar",
+					},
+					cli.BoolFlag{
+						Name:  "show-versions,v",
+						Usage: "Display versions available to upgrade to",
+					},
+					cli.StringSliceFlag{
+						Name:  "target,t",
+						Usage: "Target project ids to upgrade, leave it empty to keep current targets",
```
dramich: This comment needs some clarity, if you drop a target on upgrade it's gone, it doesn't mean it's running the old version still. If I have target1 and target2 then say `upgrade --target target1` target2 gets dropped completely.

cmd/multiclusterapp.go
```diff
@@ -0,0 +1,703 @@
...
+
+	app := &managementClient.MultiClusterApp{
+		Name: appName,
+	}
+
+	resource, err := Lookup(c, templateName, managementClient.TemplateType)
+	if err != nil {
+		return err
+	}
+	template, err := c.ManagementClient.Template.ByID(resource.ID)
```
dramich: Check this error

cmd/multiclusterapp.go
```diff
@@ -0,0 +1,703 @@
...
+	}
+
+	interactive := !ctx.Bool("no-prompt")
+	answers := make(map[string]string)
+	err = processAnswers(ctx, c, templateVersion, answers, interactive)
+	if err != nil {
+		return err
+	}
+
+	targets := ctx.StringSlice("target")
```
dramich: Lookup can be used to get projects by name, this would be consistent with other resources and also the UI

cmd/util_ref.go
```diff
@@ -0,0 +1,18 @@
+package cmd
```
dramich: Drop this file and move the helpers to cmd/multiclusterapp.go

cmd/util_ref.go
```diff
@@ -0,0 +1,18 @@
...
+
+import (
+	"fmt"
+	"strings"
+)
+
+func fromStrings(namespace, name string) string {
```
dramich: Name needs to be changed to reflect what this function does, as do the arg names as this seems to be used to combine a cluster:project

cmd/util_ref.go
```diff
@@ -0,0 +1,18 @@
...
+	return fmt.Sprintf("%s:%s", namespace, name)
+}
+
+func parse(ref string) (namespace string, name string) {
```
dramich: Name needs to be changed to reflect what this function does e.g. `parseProjectID`

main.go
```diff
@@ -84,6 +84,7 @@ func mainErr() error {
 	}
 	app.Commands = []cli.Command{
 		cmd.AppCommand(),
+		cmd.MultiClusterAppCommand(),
```
dramich: This should be alphabetical

dramich:
The commands should be able to work with both versions of the target - <clusterID>:<projectID> and the happy version <clusterName>:<projectName> as this is seen in both the cli and the UI. If a command accepts one it needs to accept the other. This will require parsing the target, determine if it's happy version of ID version and proceeding from there.

cmd/multiclusterapp.go
```diff
@@ -0,0 +1,703 @@
...
+
+	defer writer.Close()
+
+	clusters, projects, err := getClusterProjectMap(ctx, c.ManagementClient)
+	if err != nil {
+		return err
+	}
+
+	for _, item := range collection.Data {
+		templateVersion, err := c.ManagementClient.TemplateVersion.ByID(item.TemplateVersionID)
```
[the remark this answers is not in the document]
gitlawr: We cannot parse templateVersionID to get the version. The ID is in the form of `cattle-global-data:library-mysql-0.3.8` but the dash separator is a valid character in template name and version field. So given `cattle-global-data:mycatalog-template-ab-cd-ef` we cannot tell whether it is `template-ab` in version `cd-ef` or `template-ab-cd` in version `ef`.

cmd/util_ref.go
gitlawr: It is also used for concatenating scope and answer key so I renamed it to `concatScope`

cmd/multiclusterapp.go

cmd/multiclusterapp.go
```diff
@@ -0,0 +1,792 @@
...
+
+func getReadableTargetNames(c *cliclient.MasterClient, clusterCache map[string]*managementClient.Cluster, projectCache map[string]*managementClient.Project, targets []managementClient.Target) ([]string, error) {
+	var targetNames []string
+	var err error
+	for _, target := range targets {
+		projectID := target.ProjectID
+		clusterID, _ := parseScope(projectID)
+		cluster, ok := clusterCache[clusterID]
+		if !ok {
+			cluster, err = c.ManagementClient.Cluster.ByID(clusterID)
```
dramich: This needs to go back to the list but also include pagination, [here](https://github.com/superseb/cli/blob/3b590c5dc4e6c369bd4119aaa576c4c6e9a8cb3a/cmd/catalog.go#L455) is an example of what that looks like. The list is going to be more efficient so that should be used for the cluster and projects.

cmd/multiclusterapp.go
```diff
@@ -0,0 +1,792 @@
...
+
+	templateVersionCache := map[string]*managementClient.TemplateVersion{}
+	clusterCache := map[string]*managementClient.Cluster{}
+	projectCache := map[string]*managementClient.Project{}
+	for _, item := range collection.Data {
+		version, err := getTemplateVersion(c.ManagementClient, templateVersionCache, item.TemplateVersionID)
+		if err != nil {
+			return err
+		}
+		targetNames, err := getReadableTargetNames(c, clusterCache, projectCache, item.Targets)
```
dramich: If getReadableTargetNames returns an error lets just use the targets that already exist on the app vs returning the error.

cmd/multiclusterapp.go
gitlawr: Updated to use cache in paginated lists.

cmd/multiclusterapp.go
gitlawr: Updated to show ids when names are not applicable

gitlawr:
@dramich Addressed comments.

cmd/multiclusterapp.go
```diff
@@ -0,0 +1,854 @@
...
+}
+
+func listAllClusters(ctx *cli.Context, client *managementClient.Client) ([]managementClient.Cluster, error) {
+	clusterCollection, err := client.Cluster.List(defaultListOpts(ctx))
+	if err != nil {
+		return nil, err
+	}
+	clusterCollectionData := clusterCollection.Data
+	for {
+		clusterCollection, _ = clusterCollection.Next()
```
dramich: Errors should not be thrown away, this needs to be checked.

cmd/multiclusterapp.go
```diff
@@ -0,0 +1,854 @@
...
+}
+
+func listAllProjects(ctx *cli.Context, client *managementClient.Client) ([]managementClient.Project, error) {
+	projectCollection, err := client.Project.List(defaultListOpts(ctx))
+	if err != nil {
+		return nil, err
+	}
+	projectCollectionData := projectCollection.Data
+	for {
+		projectCollection, _ = projectCollection.Next()
```
dramich: Errors should not be thrown away, this needs to be checked.

cmd/multiclusterapp.go
```diff
@@ -0,0 +1,854 @@
...
+		if !projectCollection.Pagination.Partial {
+			break
+		}
+	}
+	return projectCollectionData, nil
+}
+
+func getReadableTargetNames(clusterCache map[string]managementClient.Cluster, projectCache map[string]managementClient.Project, targets []managementClient.Target) []string {
+	var targetNames []string
+	var err error
```
dramich: This is never set to anything so it's always Nil, remove if not needed.

cmd/multiclusterapp.go
```diff
@@ -0,0 +1,854 @@
...
+	for _, target := range targets {
+		projectID := target.ProjectID
+		clusterID, _ := parseScope(projectID)
+		cluster, ok := clusterCache[clusterID]
+		if !ok {
+			logrus.Debugf("Cannot get readable name for target %q: %v\nShow the target in ID instead.", target.ProjectID, err)
```
dramich: Drop `err` from this and the newline, try to keep logs to a single line. `"Cannot get readable name for target %q, showing ID"`

cmd/multiclusterapp.go
```diff
@@ -0,0 +1,854 @@
...
+			targetNames = append(targetNames, target.ProjectID)
+			continue
+		}
+		project, ok := projectCache[projectID]
+		if !ok {
+			logrus.Debugf("Cannot get readable name for target %q: %v\nShow the target in ID instead.", target.ProjectID, err)
```
dramich: Drop `err` from this and the newline, try to keep logs to a single line. `"Cannot get readable name for target %q, showing ID"`

cmd/multiclusterapp.go
```diff
@@ -0,0 +1,854 @@
...
+			return "", err
+		}
+		templateVersionCache[templateVersion.ID] = templateVersion.Version
+		version = templateVersion.Version
+	}
+	return version, nil
+}
+
+func getClusterProjectMap(ctx *cli.Context, client *managementClient.Client) (map[string]managementClient.Cluster, map[string]managementClient.Project, error) {
+	clusters := map[string]managementClient.Cluster{}
```
dramich: Instantiate with `make`

cmd/multiclusterapp.go
```diff
@@ -0,0 +1,854 @@
...
+	clusterCollectionData, err := listAllClusters(ctx, client)
+	if err != nil {
+		return nil, nil, err
+	}
+	for _, c := range clusterCollectionData {
+		clusters[c.ID] = c
+	}
+	projects := map[string]managementClient.Project{}
```
dramich: Instantiate with `make`

gitlawr:
@dramich Updated

dramich:
LGTM

gitlawr:
Squashed

alena1108:
@gitlawr : there were some RBAC related changes that got merged to rancher just recently: https://github.com/rancher/rancher/pull/17721. Some MCA updates got moved into actions vs being accepted as PUT. @mrajashree is going to follow up with you on the details. We can either incorporate the changes to this pr, or to the follow up one.

alena1108:
Follow up: will merge the pr as it is for now. Addressing RBAC changes are tracked in https://github.com/rancher/rancher/issues/17981 and https://github.com/rancher/rancher/issues/17982

PR #216, opened 2019-01-22
CLI no longer returns a 404 error upon issuing project ls command.

rmweir:
Problem: CLI returns 404 when issuing "rancher project ls" command after toggle rancher project "add-local" argument from true to false.

Solution: Set cli2.json project key's value to empty, triggering expected warning instead.

Issue: https://github.com/rancher/rancher/issues/16981

### DIFF
--- a/cmd/app.go
+++ b/cmd/app.go
@@ -31,8 +31,8 @@ const (
 	installAppDescription = `
 Install an app template in the current Rancher server. This defaults to the newest version of the app template.
 Specify a version using '--version' if required.
-The app will be installed into a new namespace unless '--namespace' is specified. 
-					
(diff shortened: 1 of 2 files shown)


cmd/project.go
```diff
@@ -107,7 +108,21 @@ func ProjectCommand() cli.Command {
 func projectLs(ctx *cli.Context) error {
 	c, err := GetClient(ctx)
 	if err != nil {
-		return err
+		path := ctx.GlobalString("cf")
```
dramich: The fix should be moved into GetClient as this is going to affect anything that calls it. The fix should not rewrite the config file as we don't know what is causing the 404. The error coming back should be checked and if it is a 404 from the project, extend the error and provide info on how to select another project - either context switch if possible or login

PR #219, opened 2019-01-23
Fix project level upgrade

gitlawr:
Address issue: https://github.com/rancher/rancher/issues/17511

Problem:
Cannot upgrade an app deployed from project level catalog. When constructing the externalId filter, catalog id gets url-encoded twice, therefore fails to get the templateversion.

Solution:
Construct the query properly.


loganhz:
@orangedeng Please help to review it

cmd/app.go
```diff
@@ -352,6 +348,22 @@ func appUpgrade(ctx *cli.Context) error {
...
 
+func getExternalIDInVersion(externalID string, version string) (string, error) {
+	u, err := url.Parse(externalID)
+	if err != nil {
+		return "", err
+	}
+
+	q := u.Query()
+	q.Set("version", version)
+	// Don't need encoding here because it as a whole is the value of the externalId filter
```
orangedeng: This comment is confusing. It says that don't encode here but we got `q.Encode()` in the next line.

gitlawr: Updated

orangedeng:
LGTM

cjellick:
@dramich can you please review
@mrajashree can you review as well

cmd/app.go
```diff
@@ -352,6 +348,22 @@ func appUpgrade(ctx *cli.Context) error {
 	return c.ProjectClient.App.ActionUpgrade(app, au)
 }
 
+func getExternalIDInVersion(externalID string, version string) (string, error) {
```
dramich: Nit: This function name should change to something like `updateExternalIDVersion`

cmd/app.go
```diff
@@ -352,6 +348,22 @@ func appUpgrade(ctx *cli.Context) error {
...
+	u, err := url.Parse(externalID)
+	if err != nil {
+		return "", err
+	}
+
+	q := u.Query()
+	q.Set("version", version)
+	// Need to unescape here because it as a whole is the value of the externalId filter
```
dramich: `// Need to unescape here to get the raw string for the externalId filter`

cmd/app_test.go
```diff
@@ -55,43 +57,34 @@ configmap: |-
...
+	assert.Equal("FLUSHDB,FLUSHALL", answers["master.disableCommands"], "unexpected master.disableCommands")
+	assert.Equal("", answers["master.service.loadBalancerIP"], "unexpected master.service.loadBalancerIP")
+	assert.Equal("ReadWriteOnce", answers["master.persistence.accessModes[0]"], "unexpected master.persistence.accessModes[0]")
+	assert.Equal("# Redis configuration file\nbind 127.0.0.1\nport 6379", answers["configmap"], "unexpected configmap")
+}
+
+func TestGetExternalIDInVersion(t *testing.T) {
+	assert := assert.New(t)
+
+	got, _ := getExternalIDInVersion("catalog://?catalog=library&template=cert-manager&version=v0.5.2", "v1.2.3")
```
dramich: This should not throw away errors, they need to be checked.

cmd/app_test.go
```diff
@@ -55,43 +57,34 @@ configmap: |-
...
+	assert.Equal(got, "catalog://?catalog=library&template=cert-manager&version=v1.2.3")
+
+	got, _ = getExternalIDInVersion("catalog://?catalog=p-826f4/pcatalog&template=mysql&version=0.3.8", "0.3.9")
```
dramich: This should not throw away errors, they need to be checked.

gitlawr:
Tested with both cluster/project catalogs.
Addressed above comments.

dramich:
LGTM

PR #220, opened 2019-01-26
Allow for local cluster when parsing context

cjellick:
"local" is a valid cluster id. Our parsing function wasn't allowing
for it.

https://github.com/rancher/rancher/issues/17647


dramich:
LGTM

PR #221, opened 2019-02-12
Fix inspect and wait

dramich:
Problem:
ByID errors on valid resources

Solution:
Return if no error to stop fallthrough to final error

Issue: https://github.com/rancher/rancher/issues/17889 and https://github.com/rancher/rancher/issues/15692

### DIFF
--- a/cliclient/cliclient.go
+++ b/cliclient/cliclient.go
@@ -2,6 +2,7 @@ package cliclient
 
 import (
 	"errors"
+	"fmt"
 	"strings"
 
 	errorsPkg "github.com/pkg/errors"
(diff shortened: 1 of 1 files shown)


cliclient/cliclient.go
```diff
@@ -148,18 +149,21 @@ func (mc *MasterClient) ByID(resource *ntypes.Resource, respObject interface{})
 		if err != nil {
 			return err
 		}
+		return nil
```
StrongMonkey: maybe `return mc.ManagementClient.ByID(resource.Type, resource.ID, &respObject)` looks better?

StrongMonkey:
LGTM

cliclient/cliclient.go
dramich: Changed it!

PR #222, opened 2019-02-13
Fix wait

dramich:
Problem:
Wait errors on already completed cluster
Error message has formatting

Solution:
Update to check status of watched object to every half second
Use value formatting for messages

Issue: https://github.com/rancher/rancher/issues/15692


StrongMonkey:
LGTM

PR #224, opened 2019-02-17
Address mcapp RBAC changes

gitlawr:
Address issue:
https://github.com/rancher/rancher/issues/17147
https://github.com/rancher/rancher/issues/17981

Related types PR: https://github.com/rancher/types/pull/726

- Add subcommand for updating target projects
- Add subcommand for updating members
- Add --member, --member-access-type option on install
- Add --role option on install/upgrade
- Add member/role details in show-app command

Examples:
```
#update targets command:
1. rancher mcapp add-project myapp mycluster:Project1 mycluster:Project2
2. rancher mcapp delete-project myapp mycluster:Project1 mycluster:Project2
# update member command:
3. rancher mcapp add-member myapp owner user1 user2
4. rancher mcapp delete-member myapp user1 user2
# set members on installation
5. rancher mcapp install --member user1 --member-access-type member --target mycluster:Default mysql myapp
# set roles on install/upgrade
6. rancher mcapp install --role project-owner --target mycluster:Default mysql myapp
7. rancher mcapp upgrade --role project-owner myapp 0.3.8
# show app command
8. rancher mcapp sa myapp
CURRENT   REVISION              CREATED
*         mcapprevision-rhcw9   18 Feb 2019 05:10:39 UTC
          mcapprevision-gmzgh   18 Feb 2019 05:10:00 UTC

CURRENT   VERSION
          0.3.7
*         0.3.8

MEMBER_NAME     MEMBER_TYPE   ACCESS_TYPE
Default Admin   Local User    owner

ROLE_NAME
cluster-owner
project-owner
```

### DIFF
--- a/cmd/multiclusterapp.go
+++ b/cmd/multiclusterapp.go
@@ -3,12 +3,14 @@ package cmd
 import (
 	"errors"
 	"fmt"
+	"net/url"
 	"sort"
 	"strings"
 	"time"
 
 	"github.com/rancher/cli/cliclient"
 	"github.com/rancher/norman/types"
+	"github.com/rancher/norman/types/slice"
 	managementClient "github.com/rancher/types/client/management/v3"
 	"github.com/sirupsen/logrus"
 	"github.com/urfave/cli"
@@ -34,10 +36,19 @@ Example:
 `
 )
 
+var memberAccessTypes = []string{"owner", "member", "read-only"}
+
 type MultiClusterAppData struct {
+	ID      string
 	App     managementClient.MultiClusterApp
 	Version string
 	Targets string
+}
+
+type MemberData struct {
+	Name       string
+	MemberType string
+	AccessType string
 }
 
 func MultiClusterAppCommand() cli.Command {
@@ -101,6 +112,19 @@ func MultiClusterAppCommand() cli.Command {
 						Name:  "target,t",
 						Usage: "Target project names/ids to install the app into",
 					},
+					cli.StringSliceFlag{
+						Name:  "role",
+						Usage: "Set roles required to launch/manage the apps in target projects",
+					},
+					cli.StringSliceFlag{
+						Name:  "member",
+						Usage: "Set members of the app, with the same access type defined by --member-access-type",
+					},
+					cli.StringFlag{
+						Name:  "member-access-type",
+						Usage: "Access type of the members. Specify only one value, and it applies to all members defined by --member. Valid options are 'owner', 'member' and 'read-only'",
+						Value: "owner",
+					},
 					cli.IntFlag{
 						Name:  "timeout",
 						Usage: "Time in seconds to wait until the app is in a ready state",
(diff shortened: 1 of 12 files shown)


cmd/multiclusterapp.go
```diff
@@ -101,6 +112,19 @@ func MultiClusterAppCommand() cli.Command {
...
+					cli.StringSliceFlag{
+						Name:  "role",
+						Usage: "Set roles required to view the app",
+					},
+					cli.StringSliceFlag{
+						Name:  "member",
+						Usage: "Set members who can view the app, with access type defined by --member-access-type",
+					},
+					cli.StringFlag{
+						Name:  "member-access-type",
```
orangedeng: Should each members have their own access type?

cmd/multiclusterapp.go
```diff
@@ -138,17 +162,59 @@ func MultiClusterAppCommand() cli.Command {
...
 			},
+			cli.Command{
+				Name:        "delete-project",
+				Usage:       "Delete target projects from a multi-cluster app",
+				Action:      deleteMcappTargetProject,
+				Description: "Examples:\n #Delete 'p1' project in cluster 'mycluster' from target projects of a multi-cluster app named 'myapp'\n rancher multiclusterapp delete-project myapp mycluster:p1\n",
+				ArgsUsage:   "[APP_NAME/APP_ID, CLUSTER_NAME:PROJECT_NAME/PROJECT_ID...]",
+			},
+			cli.Command{
+				Name:        "add-member",
```
orangedeng: Should add-member also have the member-access-type parameter?

cmd/multiclusterapp.go
gitlawr: It's not easy to claim different members with different access type in a single command. 
Therefore,
1. These arguments are for setting a set of members with the same access type on install.
2. We introduce add/delete-member to do accurate membership assignments

cmd/multiclusterapp.go
gitlawr: Access-type is included in the command args. The usage of add-member is:
```
"[APP_NAME/APP_ID, USER_NAME/USER_ID, ACCESS_TYPE]"
rancher mcapp add-member user1 read-only
```
Or do you think it's better if it goes as
```
rancher mcapp add-member --member-access-type read-only u1 u2 u3...
```

orangedeng: If so, I would prefer using rancher mcapp add-member myapp owner user1 user2 ... Put the scaleable parameters in the end then you can add multiple user in this command.

orangedeng:
LGTM

gitlawr:
Types RP is merged. Rebased.

cjellick:
@dramich to review

cmd/multiclusterapp.go
```diff
@@ -679,45 +763,203 @@ func fromMultiClusterAppAnswers(answers []managementClient.Answer) map[string]st
...
+	app, input, err := proceedUpdateMultiClusterAppTargetsArgs(ctx, c)
+	if err != nil {
+		return err
+	}
+	return c.ManagementClient.MultiClusterApp.ActionRemoveProjects(app, input)
+}
+
+func proceedUpdateMultiClusterAppTargetsArgs(ctx *cli.Context, c *cliclient.MasterClient) (*managementClient.MultiClusterApp, *managementClient.UpdateMultiClusterAppTargetsInput, error) {
+	if len(ctx.Args()) < 2 {
+		return nil, nil, cli.ShowSubcommandHelp(ctx)
```
dramich: `cli.ShowSubcommandHelp(ctx)` is not an error, it prints the help for the command, so the error checks in `addMcappTargetProject` and `deleteMcappTargetProject` don't catch it which causes a panic further down. Check the length of args in the first function.

cmd/multiclusterapp.go
```diff
@@ -679,45 +763,203 @@ func fromMultiClusterAppAnswers(answers []managementClient.Answer) map[string]st
...
+	}
+
+	for _, name := range memberNames {
+		member, err := searchForMember(ctx, c, name)
+		if err != nil {
+			return err
+		}
+
+		memberID := member.ID
+		toKeepMembers := make([]managementClient.Member, 0, len(app.Members))
```
dramich: This isn't needed. `var toKeepMembers []managementClient.Member` is fine for this.

cmd/multiclusterapp.go
```diff
@@ -803,6 +1030,46 @@ func outputMultiClusterAppRevisions(ctx *cli.Context, c *cliclient.MasterClient)
 	return writer.Err()
 }
 
+func outputMultiClusterAppMembers(ctx *cli.Context, c *cliclient.MasterClient, app *managementClient.MultiClusterApp) error {
```
dramich: I think this should be its own command `list-members` following what is done in project. The `show-app` is about version info and if there are a lot of users on that app it's going to make for a large output.

cmd/multiclusterapp.go
```diff
@@ -803,6 +1030,46 @@ func outputMultiClusterAppRevisions(ctx *cli.Context, c *cliclient.MasterClient)
...
+		writer.Write(&MemberData{
+			Name:       principal.Name,
+			MemberType: strings.Title(fmt.Sprintf("%s %s", principal.Provider, principal.PrincipalType)),
+			AccessType: m.AccessType,
+		})
+	}
+	return writer.Err()
+}
+
+func outputMultiClusterAppRoles(ctx *cli.Context, c *cliclient.MasterClient, app *managementClient.MultiClusterApp) error {
```
dramich: Hide this behind a flag on `show-app`

cmd/multiclusterapp.go
dramich: Agreed on keeping it this way, it's how project and cluster members work.

cmd/multiclusterapp.go
```diff
@@ -101,6 +112,19 @@ func MultiClusterAppCommand() cli.Command {
 						Name:  "target,t",
 						Usage: "Target project names/ids to install the app into",
 					},
+					cli.StringSliceFlag{
+						Name:  "role",
+						Usage: "Set roles required to view the app",
```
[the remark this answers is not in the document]
gitlawr: This is the same as what we display in the UI. Probably want to hide implementation details of the sa? @mrajashree Can you confirm if it's good or we need to tune the UI as well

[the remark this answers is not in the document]
mrajashree: Yes the description seems incorrect. These roles are what the system account or creator of the actual app instances will get. So maybe we can change it to "Set roles required to launch/manage the app"?

cmd/multiclusterapp.go
gitlawr: Do you expect a `list-roles` command to display roles info of a mcapp

gitlawr: nevermind, just saw your comment below

gitlawr:
@dramich Addressed above comments.

cmd/multiclusterapp.go
```diff
@@ -138,17 +162,70 @@ func MultiClusterAppCommand() cli.Command {
...
+				Usage:       "Delete members from a multi-cluster app",
+				Action:      deleteMcappMember,
+				Description: "Examples:\n #Delete the membership of a user named 'user1' from a multi-cluster app named 'myapp'\n rancher multiclusterapp delete-member myapp user1\n",
+				ArgsUsage:   "[APP_NAME/APP_ID, USER_NAME/USER_ID...]",
+			},
+			cli.Command{
+				Name:        "list-members",
+				Aliases:     []string{"lm"},
+				Usage:       "List current members of a multi-cluster app",
+				Description: "List current members of a multi-cluster app",
```
dramich: Either update description to be different or drop it, having them the same makes the help output look funny.

cmd/multiclusterapp.go
```diff
@@ -679,6 +778,156 @@ func fromMultiClusterAppAnswers(answers []managementClient.Answer) map[string]st
...
+	return input, nil
+}
+
+func addMcappMember(ctx *cli.Context) error {
+	if len(ctx.Args()) < 3 {
+		return cli.ShowSubcommandHelp(ctx)
+	}
+
+	appName := ctx.Args().First()
+	accessType := ctx.Args().Get(1)
```
dramich: Update this to use `strings.ToLower`

cmd/multiclusterapp.go
```diff
@@ -522,6 +594,33 @@ func multiClusterAppTemplateInstall(ctx *cli.Context) error {
 	}
 	app.TemplateVersionID = templateVersionID
 
+	roles := ctx.StringSlice("role")
+	if len(roles) > 0 {
+		app.Roles = roles
+	}
+
+	members := ctx.StringSlice("member")
+	accessType := ctx.String("member-access-type")
```
dramich: Update this to use `strings.ToLower` to make the comparison nicer for the user.

gitlawr:
@dramich Updated

dramich:
LGTM

cmd/multiclusterapp.go
```diff
@@ -34,12 +36,21 @@ Example:
...
+
 type MultiClusterAppData struct {
+	ID      string
 	App     managementClient.MultiClusterApp
 	Version string
 	Targets string
 }
 
+type MemberData struct {
+	Name       string
```
mrajashree: Member has different fields for groupPrincipalId and userPrincipalId https://github.com/rancher/types/blob/master/apis/management.cattle.io/v3/multi_cluster_app.go#L56
How do we differentiate between them here?

cmd/multiclusterapp.go
```diff
@@ -138,17 +162,69 @@ func MultiClusterAppCommand() cli.Command {
 						Name:  "set",
 						Usage: "Set answers for the template, can be used multiple times. Example: --set foo=bar",
 					},
+					cli.StringSliceFlag{
+						Name: "role,r",
+						Usage: "Set roles required to launch/manage the apps in target projects. Specified roles on " +
+							"upgrade will override all the original roles. Leave it empty to keep current roles",
```
mrajashree: Why are roles on upgrade overriding all existing roles? If user has to add more roles will they be expected to provide all existing roles as input again? If so then can we add that as well to the usage?

cmd/multiclusterapp.go
gitlawr: It is shown in `memberType`, See L1002:
```
MemberType: strings.Title(fmt.Sprintf("%s %s", principal.Provider, principal.PrincipalType))
```

cmd/multiclusterapp.go
gitlawr: It's a simpler way to update roles. Or we will need to provide different flags/commands to add or delete roles of a mcapp. Updated the usage.

gitlawr:
Squashed

mrajashree:
LGTM

PR #225, opened 2019-02-24
Add globaldns commands

gitlawr:
Address issue: https://github.com/rancher/rancher/issues/17982

- add globaldns provider/entry commands
- add create/update/delete/list-members/add-member/delete-member
commands for provider/entry
- add add-project/delete-project commands for entry

Example command flows:
```
$ rancher globaldns provider ls
$ rancher globaldns provider create dnsp --type route53 --root-domain test.dm
$ rancher globaldns provider update dnsp --root-domain sub.test.dm
$ rancher globaldns provider delete p1 p2
$ rancher globaldns provider list-members p1
$ rancher globaldns provider add-member p1 user1 user2
$ rancher globaldns provider delete-member p1 user1 user2

$ rancher globaldns entry ls
$ rancher globaldns entry create --fqdn test.dm --provider p1
$ rancher globaldns entry update ENTRY_ID --fqdn test.dm
$ rancher globaldns entry delete ID1 ID2
$ rancher globaldns entry list-members ID
$ rancher globaldns entry add-member ID user1 user2
$ rancher globaldns entry delete-member ID user1 user2
$ rancher globaldns entry add-project ID p1 p2
$ rancher globaldns entry delete-project ID p1 p2
```

### DIFF
--- a/cmd/globaldns.go
+++ b/cmd/globaldns.go
@@ -1,0 +1,897 @@
+package cmd
+
+import (
+	"errors"
+	"fmt"
+	"strings"
+
+	"github.com/rancher/cli/cliclient"
+	managementClient "github.com/rancher/types/client/management/v3"
+	"github.com/urfave/cli"
+)
+
+const (
+	alidnsProvider     = "alidns"
+	cloudflareProvider = "cloudflare"
+	route53Provider    = "route53"
+
+	envAlibabaCloudAccessKey = "ALI_ACCESS_KEY_ID"
+	envAlibabaCloudSecretKey = "ALI_ACCESS_KEY_SECRET"
+	envAwsAccessKey          = "AWS_ACCESS_KEY_ID"
+	envAwsSecretKey          = "AWS_SECRET_ACCESS_KEY"
+	envCloudflareAPIEmail    = "CF_API_EMAIL"
+	envCloudflareAPIKey      = "CF_API_KEY"
+
+	argAlibabaCloudAccessKey = "alibabacloud-access-key-id"
+	argAlibabaCloudSecretKey = "alibabacloud-access-key-secret"
+	argAwsAccessKey          = "aws-access-key"
+	argAwsSecretKey          = "aws-secret-key"
+	argCloudflareAPIEmail    = "cloudflare-api-email"
+	argCloudflareAPIKey      = "cloudflare-api-key"
+
+	argFQDN            = "fqdn"
+	argMember          = "member"
+	argMultiClusterApp = "multi-cluster-app"
+	argProject         = "project"
+	argProvider        = "provider"
+	argRootDomain      = "root-domain"
+	argTTL             = "ttl"
+	argType            = "type"
+
+	memberAccessTypeOwner = "owner"
+)
+
+type globalDNSHolder struct {
+	ID       string
+	Provider managementClient.GlobalDNSProvider
+	Entry    managementClient.GlobalDNS
+	Target   string
+}
+
+var globalDNSProviderCredentialFlags = []cli.Flag{
+	cli.StringFlag{
+		Name:   argAlibabaCloudAccessKey,
+		EnvVar: envAlibabaCloudAccessKey,
+		Usage:  "Alibaba Cloud access key ID for alidns provider",
+	},
+	cli.StringFlag{
+		Name:   argAlibabaCloudSecretKey,
+		EnvVar: envAlibabaCloudSecretKey,
+		Usage:  "Alibaba Cloud access key secret for alidns provider",
+	},
+	cli.StringFlag{
+		Name:   argAwsAccessKey,
+		EnvVar: envAwsAccessKey,
+		Usage:  "AWS access key for route53 provider",
+	},
+	cli.StringFlag{
+		Name:   argAwsSecretKey,
+		EnvVar: envAwsSecretKey,
+		Usage:  "AWS secret key for route53 provider",
+	},
+	cli.StringFlag{
+		Name:   argCloudflareAPIEmail,
+		EnvVar: envCloudflareAPIEmail,
+		Usage:  "API email for Cloudflare provider",
+	},
+	cli.StringFlag{
+		Name:   argCloudflareAPIKey,
+		EnvVar: envCloudflareAPIKey,
+		Usage:  "API key for Cloudflare provider",
+	},
+}
+
+func GlobalDNSCommand() cli.Command {
+	return cli.Command{
+		Name:  "globaldns",
+		Usage: "Operations on global DNS providers and entries",
+		Subcommands: []cli.Command{
+			{
+				Name:    "providers",
+				Aliases: []string{"provider"},
+				Usage:   "Operations on global DNS providers",
+				Action:  defaultAction(globalDNSProviderLs),
+				Subcommands: []cli.Command{
+					{
+						Name:   "ls",
+						Usage:  "List global DNS providers",
+						Action: globalDNSProviderLs,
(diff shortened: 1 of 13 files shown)


cmd/globaldns.go
```diff
@@ -0,0 +1,865 @@
...
+		providerType = route53Provider
+	} else if provider.CloudflareProviderConfig != nil {
+		providerType = cloudflareProvider
+	} else if provider.AlidnsProviderConfig != nil {
+		providerType = alidnsProvider
+	}
+
+	switch providerType {
+	case route53Provider:
+		accessKey := os.Getenv(envAwsAccessKeyID)
```
orangedeng: Should we have a way to get those parameters from cli flags instead of getting them from environment variable?

orangedeng: like we can support --args AWS_ACCESS_KEY_ID=xxxxxxx --args AWS_SECRET_ACCESS_KEY=xxxx

cmd/globaldns.go
```diff
@@ -0,0 +1,865 @@
...
+
+	if ctx.IsSet(argProvider) {
+		globalDNSProvider, err := searchForGlobalDNSProvider(c, ctx.String(argProvider))
+		if err != nil {
+			return err
+		}
+		update["providerId"] = globalDNSProvider.ID
+	}
+
+	if ctx.IsSet(argMultiClusterApp) {
```
orangedeng: I think you need to validate that we can not support both `argProvider` and `argMultiClusterApp `  target in one update command.

gitlawr: Why can't we do that?

orangedeng: nvm, I see it now,

cmd/globaldns.go
gitlawr: Updated

orangedeng:
LGTM

cmd/globaldns.go
```diff
@@ -0,0 +1,931 @@
...
+	envAwsSecretKey          = "AWS_SECRET_ACCESS_KEY"
+	envCloudflareAPIEmail    = "CF_API_EMAIL"
+	envCloudflareAPIKey      = "CF_API_KEY"
+
+	argAwsAccessKey          = "aws-access-key"
+	argAwsSecretKey          = "aws-secret-key"
+	argCloudflareAPIEmail    = "cloudflare-api-email"
+	argCloudflareAPIKey      = "cloudflare-api-key"
+	argAlibabaCloudAccessKey = "alibabacloud-access-key-id"
+	argAlibabaCloudSecretKey = "alibabacloud-access-key-secret"
```
dramich: Nit: Make alphabetical

cmd/globaldns.go
```diff
@@ -0,0 +1,931 @@
...
+)
+
+const (
+	alidnsProvider     = "alidns"
+	cloudflareProvider = "cloudflare"
+	route53Provider    = "route53"
+
+	envAlibabaCloudAccessKey = "ALI_ACCESS_KEY_ID"
+	envAlibabaCloudSecretKey = "ALI_ACCESS_KEY_SECRET"
+	envAwsAccessKey          = "AWS_ACCESS_KEY_ID"
```
dramich: Make these match the UI  e.g. `AWS_ACCESS_KEY`

cmd/globaldns.go
```diff
@@ -0,0 +1,931 @@
...
+		return cli.ShowSubcommandHelp(ctx)
+	}
+
+	c, err := GetClient(ctx)
+	if err != nil {
+		return err
+	}
+
+	name := ctx.Args().First()
+	providerType := ctx.String(argType)
```
dramich: strings.ToLower this

cmd/globaldns.go
```diff
@@ -0,0 +1,931 @@
...
+							},
+							cli.StringFlag{
+								Name:  argRootDomain,
+								Usage: "Set root domain of a global DNS provider",
+							},
+							cli.StringSliceFlag{
+								Name:  argMember,
+								Usage: "Set members of the global DNS provider, can be used multiple times",
+							},
+							cli.StringFlag{
```
dramich: Put the 6 key flags together outside the command creation then add the reference like it's done with formatFlag and quietFlag.

cmd/globaldns.go
```diff
@@ -0,0 +1,931 @@
...
+}
+
+func getEntryTarget(c *cliclient.MasterClient, clusterCache map[string]managementClient.Cluster, projectCache map[string]managementClient.Project, entry *managementClient.GlobalDNS) (string, error) {
+	target := ""
+	if entry.MultiClusterAppID != "" {
+		_, app, err := searchForMcapp(c, entry.MultiClusterAppID)
+		if err != nil {
+			return "", err
+		}
+		target = fmt.Sprintf("Multi Cluster App %s", app.Name)
```
dramich: Update to `Multi-Cluster App: %s`

cmd/globaldns.go
```diff
@@ -0,0 +1,931 @@
...
+								Usage: "Set projects as the target to which a global DNS entry resolves, can be used multiple times",
+							},
+							cli.StringSliceFlag{
+								Name:  argMember,
+								Usage: "Set members of a global DNS entry, can be used multiple times",
+							},
+						},
+					},
+					{
+						Name:      "update",
```
dramich: Why doesn't update support projects as a flag as well? Keeping the separate command for projects is fine.

cmd/globaldns.go
```diff
@@ -0,0 +1,931 @@
...
+		return err
+	}
+
+	update := make(map[string]interface{})
+
+	if ctx.IsSet(argRootDomain) {
+		update["rootDomain"] = ctx.String(argRootDomain)
+	}
+
+	providerType := ""
```
dramich: var providerType string

cmd/globaldns.go
```diff
@@ -0,0 +1,931 @@
...
+			ID:     entry.ID,
+			Entry:  entry,
+			Target: target,
+		})
+	}
+	return writer.Err()
+}
+
+func getEntryTarget(c *cliclient.MasterClient, clusterCache map[string]managementClient.Cluster, projectCache map[string]managementClient.Project, entry *managementClient.GlobalDNS) (string, error) {
+	target := ""
```
dramich: var target string

cmd/globaldns.go
```diff
@@ -0,0 +1,931 @@
...
+	}
+
+	update := make(map[string]interface{})
+
+	if ctx.IsSet(argRootDomain) {
+		update["rootDomain"] = ctx.String(argRootDomain)
+	}
+
+	providerType := ""
+	if provider.Route53ProviderConfig != nil {
```
dramich: Pull the switch up into these if blocks with an `else` for the error, `providerType` isn't used for anything else so no reason to check the same thing twice.

cmd/globaldns.go
gitlawr: Project target is `noupdate` fields in the API and it needs to be done via addProjects/removeProjects actions

cmd/globaldns.go
gitlawr: Where can I see `AWS_ACCESS_KEY` in the UI? I use this because it is commonly used by aws cli/sdk

gitlawr:
@dramich Updated

cmd/globaldns.go
dramich: That makes sense, if that is how the aws cli works leave it.

gitlawr:
Squashed

PR #227, opened 2019-03-07
Fix panic on mcapp install

gitlawr:
Address issue: https://github.com/rancher/rancher/issues/18670

Problem:
Panic happens on mcapp install when talking to a 2C4G single node K8S with Rancher HA installed and a GKE cluster provisioned. It does not happen in a standalone rancher. 
Inspecting the code, it is probably because the heavy loaded Rancher/K8S takes longer time to handle the conditions thus we get nil status when fetching the mcapp object after it is created.

### DIFF
--- a/cmd/multiclusterapp.go
+++ b/cmd/multiclusterapp.go
@@ -622,6 +622,10 @@ func multiClusterAppTemplateInstall(ctx *cli.Context) error {
 		if err != nil {
 			return err
 		}
+		if madeApp.Status == nil {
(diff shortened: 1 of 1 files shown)


orangedeng:
I don't think it is a good way to fix it but we can still have this PR for workaround. I would like to have a comment for why you are doing this in the code and we can fix it later.

gitlawr:
The mcapp.status is initialized with nil, so we should do the check anyway. The panic does not happen when the controller updates status before the client gets the mcapp object.
Added a comment for explanation.

orangedeng:
LGTM

PR #228, opened 2019-03-08
Add current environment and add KUBECONFIG as config location

BobVanB:
**Problem:**
In de default manual of kubernetes.io it is possible to extend the kubectl binary.
If i have a simple plugin named `kubectl-foo` in my custom PATH variable like so:
```bash
#!/bin/bash
echo "I am a plugin named kubectl-foo"
```
This will work 
```
$ kubectl foo
I am a plugin named kubectl-foo
```
when i pass it to rancher
```
rancher kubectl foo
Error: unknown command "foo" for "kubectl"
```
**Analyzed problem:**
`--kubeconfig` is passed to kubectl
If i experiment a little with what the kubectl does. I noticed that kubectl ignores all global parameters and leaves this to the plugin. Witch means that the following command will fail.
```
$ kubectl --kubeconfig=/tmp/rancher-test foo
Error: unknown command "foo" for "kubectl"
```
This would work:
```
$ kubectl foo --kubeconfig=/tmp/rancher-test
I am a plugin named kubectl-foo
```
Here the problem is that `--kubeconfig=...` is passed to the foo plugin,
and not used by kubectl. But it works. Witch leaves the parsing of --kubeconfig to the foo plugin.
And we never have to look at it again.

One other solution can be to set the kubeconfig as an environment variable.
This will give the plugin direct access to kubectl commands.
Like this plugin:
```
#!/bin/bash
kubectl get nodes
```
```
Result:
$ rancher kubectl foo
NAME   STATUS   ROLES               AGE   VERSION
cp01   Ready    controlplane,etcd   17d   v1.11.6
w01    Ready    worker              17d   v1.11.6
```
**Possible solutions**
1. append the `--kubeconfig=` to the command and leave the handling of the parameter to the config file.
2. append/overwrite any KUBECONFIG environment variable, and give default access for further kubectl commands.

With option 2 it is always possible to still pass the `--kubeconfig` and overwrite `KUBECONFIG`, that decision is left to the plugin itself.

**Final Words:**
I want to add addons to the rancher cli, this way i only need one cli entrypoint.
I do know that starting with `--kubeconfig` for plugins is broken.

**Change Notes**
I have to pass `os.Environ()`, if i only add `KUBECONFIG` to `cmd.Env`, all the other environment values are gone.

With kind regards.


maggieliu:
related to: https://github.com/rancher/rancher/issues/26992

PR #229, opened 2019-03-12
Add roles in add/delete-member API input

gitlawr:
Address https://github.com/rancher/rancher/issues/17981#issuecomment-471649787

Problem:
When using add-member/delete-member commands, missing roles on updating mcapp
is rejected on API layer.

Solution:
Add existing roles to the input.


orangedeng:
LGTM

PR #230, opened 2019-03-12
Mcapp enhancements

gitlawr:
Address https://github.com/rancher/rancher/issues/17147#issuecomment-471712299

1. Add `list-answers` command to print current answers of a mcapp
```
$ rancher mcapp list-answers myapp
SCOPE                            QUESTION              ANSWER
Global                           mysqlDatabase         admin
Global                           mysqlPassword         abc000
Global                           mysqlUser             admin
Global                           persistence.enabled   false
Global                           service.port          3306
Global                           service.type          ClusterIP
Global                           defaultImage          true
All projects in cluster lawr-t   mysqlPassword         abc222
Project Default                  mysqlPassword         abc111
```
2. Add upgrade strategy options to install/upgrade
3. Error out on invalid version and improve checks
4. Wait for mcapp `active` state on install/upgrade/rollback/add-project and add `--timeout` option
5. Add `--rollback` option for upgrade to support rollback on failure

### DIFF
--- a/cmd/multiclusterapp.go
+++ b/cmd/multiclusterapp.go
@@ -33,9 +33,17 @@ Example:
 	# Install the redis template and set target projects to install
 	$ rancher multiclusterapp install --target mycluster:Default --target c-98pjr:p-w6c5f redis appFoo
 `
+	upgradeStrategySimultaneously = "simultaneously"
+	upgradeStrategyRollingUpdate  = "rolling-update"
+	argUpgradeStrategy            = "upgrade-strategy"
+	argUpgradeBatchSize           = "upgrade-batch-size"
+	argUpgradeBatchInterval       = "upgrade-batch-interval"
 )
 
-var memberAccessTypes = []string{"owner", "member", "read-only"}
+var (
+	memberAccessTypes = []string{"owner", "member", "read-only"}
+	upgradeStrategies = []string{upgradeStrategySimultaneously, upgradeStrategyRollingUpdate}
+)
 
 type MultiClusterAppData struct {
 	ID      string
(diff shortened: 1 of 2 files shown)


cmd/multiclusterapp.go
```diff
@@ -611,32 +734,25 @@ func multiClusterAppTemplateInstall(ctx *cli.Context) error {
...
+}
+
+func waitUntilMultiClusterAppActive(c *cliclient.MasterClient, appID string, timeout int) error {
+	var timeWait int
+	for {
+		if timeWait*2 >= timeout {
 			return errors.New("timed out waiting for app to be active, the app could still be installing. Run 'rancher multiclusterapps' to verify")
 		}
-		timewait++
+		timeWait++
```
orangedeng: Can we do `context.Timeout` and `time.Ticker` here? It would be better than we are using an endless loop here since we are changing this.

cmd/multiclusterapp.go
```diff
@@ -120,11 +135,22 @@ func MultiClusterAppCommand() cli.Command {
 						Usage: "Access type of the members. Specify only one value, and it applies to all members defined by --member. Valid options are 'owner', 'member' and 'read-only'",
 						Value: "owner",
 					},
-					cli.IntFlag{
-						Name:  "timeout",
-						Usage: "Time in seconds to wait until the app is in a ready state",
-						Value: 60,
+					cli.StringFlag{
+						Name:  argUpgradeStrategy,
```
orangedeng: I don't understand why we have upgrade strategy in install command. Can you explain a little?

gitlawr: Upgrade strategy is an attribute of multiclusterapp. It's not dedicated for a single upgrade.

cmd/multiclusterapp.go

orangedeng:
LGTM

PR #231, opened 2019-03-13
Change global DNS target in a straightforward way

gitlawr:
Address https://github.com/rancher/rancher/issues/17982#issuecomment-472225106

Problem:
Cannot change globalDNS target to mcapp when it aims at projects, and vice versa

Solution:
Delete existing project target when mcapp target is set,
Delete existing mcapp target when try to add project target.

### DIFF
--- a/cmd/globaldns.go
+++ b/cmd/globaldns.go
@@ -734,14 +734,33 @@ func globalDNSUpdate(ctx *cli.Context) error {
 			return err
 		}
 		update["multiClusterAppId"] = resource.ID
+		if len(entry.ProjectIDs) > 0 {
+			input := &managementClient.UpdateGlobalDNSTargetsInput{
+				ProjectIDs: entry.ProjectIDs,
+			}
+			if err := c.ManagementClient.GlobalDNS.ActionRemoveProjects(entry, input); err != nil {
+				return err
+			}
+		}
 	}
 
 	if ctx.IsSet(argTTL) {
 		update["ttl"] = ctx.Int64(argTTL)
 	}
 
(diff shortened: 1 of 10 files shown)


orangedeng:
LGTM

cmd/globaldns.go
```diff
@@ -734,6 +734,14 @@ func globalDNSUpdate(ctx *cli.Context) error {
 			return err
 		}
 		update["multiClusterAppId"] = resource.ID
+		if len(entry.ProjectIDs) > 0 {
+			input := &managementClient.UpdateGlobalDNSTargetsInput{
+				ProjectIDs: entry.ProjectIDs,
+			}
+			if err := c.ManagementClient.GlobalDNS.ActionRemoveProjects(entry, input); err != nil {
+				return err
```
prachidamle: What if the update fails, we should rollback to the previous state and set back the Projects?

cmd/globaldns.go
```diff
@@ -870,9 +878,16 @@ func addGlobalDNSProjects(ctx *cli.Context) error {
...
+		}
+	}
+
 	input := &managementClient.UpdateGlobalDNSTargetsInput{
-		MultiClusterAppID: "",
-		ProjectIDs:        projectIDs,
+		ProjectIDs: projectIDs,
 	}
 
 	return c.ManagementClient.GlobalDNS.ActionAddProjects(entry, input)
```
prachidamle: Same as above, handle error if ActionAddProjects fails?

gitlawr:
@prachidamle Updated

prachidamle:
thanks @gitlawr, that change looks correct

PR #235, opened 2019-06-10
Reset answers when set flag is used

rmweir:
Problem:
Using set flag adds new answers to previous answers. This does not match the behavior of helm, which first resets existing answers.

Solution:
Delete existing answers

Issue:
https://github.com/rancher/rancher/issues/17540
Also fixes:
https://github.com/rancher/rancher/issues/17405


dramich:
LGTM

cjellick:
lgtm

PR #236, opened 2019-07-11
fix yaml keys

leodotcloud:
Addresses: 
https://github.com/rancher/rancher/issues/15098


cmd/common.go
```diff
@@ -8,6 +8,7 @@ import (
 	"encoding/json"
 	"encoding/pem"
 	"fmt"
+	"github.com/rancher/norman/types/convert"
```
dramich: Needs to be moved down to imported packages.

dramich:
Small formatting thing but logic looks good

leodotcloud:
Thanks, @dramich. Fixed the formatting issue.

PR #237, opened 2019-07-18
Wait timeout fix

rmweir:
Problem:
Some apps require wait and timeout flags to be passed to helm. CLI does not support these flags for apps.

Solution:
Accept flags for wait and timeout and then pass them to helm.

Depends on following PRs:
types - https://github.com/rancher/types/pull/911
rancher - https://github.com/rancher/rancher/pull/21604

intended merge order:
types -> rancher -> cli

Issue:
https://github.com/rancher/rancher/issues/17471

### DIFF
--- a/cmd/app.go
+++ b/cmd/app.go
@@ -177,6 +177,15 @@ func AppCommand() cli.Command {
 						Name:  "no-prompt",
 						Usage: "Suppress asking questions and use the default values when required answers are not provided",
 					},
+					cli.IntFlag{
+						Name:  "timeout",
+						Usage: "Amount of time to wait for k8s commands (default is 300 secs). Example: --timeout 600",
+						Value: 300,
+					},
+					cli.BoolFlag{
+						Name:  "wait",
+						Usage: "Wait, as long as timeout value, for installed resources to be ready (pods, PVCs, deployments, etc.). Example: --wait",
+					},
 				},
 			},
 			cli.Command{
@@ -252,6 +261,12 @@ func AppCommand() cli.Command {
 				Flags: []cli.Flag{
 					formatFlag,
 				},
+			},
+			cli.Command{
+				Name:      "show-notes",
+				Usage:     "Show contents of apps notes.txt",
+				Action:    appNotes,
+				ArgsUsage: "[APP_NAME/APP_ID]",
 			},
 		},
 	}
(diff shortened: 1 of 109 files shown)


vendor.conf
```diff
@@ -28,4 +28,4 @@ github.com/pmezard/go-difflib v1.0.0
 github.com/stretchr/testify v1.2.2
 
 github.com/rancher/norman 457c15b94acae52afb5290aa315452c7621d452a
-github.com/rancher/types ee87bbdceacff5c566f901bdc57eec8330e7ed13
+github.com/rancher/types 336da48888a366ef68b859845585dc1b359a3742 https://github.com/rmweir/types
```
cbron: This is temporary for testing right ? Wouldn't think we want a personal repo on here.

cmd/app.go
```diff
@@ -658,43 +673,50 @@ func templateInstall(ctx *cli.Context) error {
...
-			if condType == "installed" && condStatus == "true" {
-				installed = true
-				break
-			}
-		}
-	}
-
 	if len(madeApp.Notes) != 0 {
 		fmt.Println(madeApp.Notes)
 	}
```
cbron: Add note about how to print notes, rather than notes if they are available.

cmd/app.go
```diff
@@ -177,6 +177,15 @@ func AppCommand() cli.Command {
 						Name:  "no-prompt",
```
cbron: Could add an example in `description` that includes these flags

cmd/app.go
```diff
@@ -658,40 +673,45 @@ func templateInstall(ctx *cli.Context) error {
...
-		}
-		for _, condition := range madeApp.Conditions {
-			condType := strings.ToLower(condition.Type)
-			condStatus := strings.ToLower(condition.Status)
-			if condType == "installed" && condStatus == "true" {
-				installed = true
-				break
-			}
-		}
+	fmt.Printf("run \"app show-notes %s\" to view app notes once app is ready", madeApp.Name)
```
cbron: Tested and it works, looks good except that I think it should have newline at end. If there is ever another print statement below it then both would be on same line. Current output: 
`run "app show-notes a-gjlhj" to view app notes once app is ready%`

cbron: fixed

cmd/app.go
```diff
@@ -658,40 +673,45 @@ func templateInstall(ctx *cli.Context) error {
 		app.TargetNamespace = namespace
 	}
 
+	app.Wait = ctx.Bool("wait")
```
luthermonson: should wait boolflag have a default of true? false? what happens if you set app.Wait to nil or whatever comes through from the not setting BoolFlag Value?

rmweir: wait a regular bool with default false, which is intended. I tested without using wait.

luthermonson: then we're good

luthermonson:
lgtm

cmd/app.go
```diff
@@ -658,40 +673,45 @@ func templateInstall(ctx *cli.Context) error {
...
+	// 300 seconds is the default value that helm already uses
+	if timeout := ctx.Int64("timeout"); timeout != 300 {
```
dramich: You shouldn't need to do the check, just use whatever timeout is passed in. If helm decides it wants to change its default to 350 this would need to be updated.

cmd/app.go
```diff
@@ -658,40 +673,45 @@ func templateInstall(ctx *cli.Context) error {
...
+	if err != nil {
+		return err
 	}
 
+	app, err := c.ProjectClient.App.ByID(resource.ID)
+	if err != nil {
+		return err
+	}
+
+	fmt.Println(app.Notes)
```
dramich: do a len check like before and print no notes or something if they don't exist. Printing a blank line will confuse people.

dramich:
lgtm

PR #239, opened 2019-07-23
Fix project-scoped catalog app upgrade

gitlawr:
Address issue: https://github.com/rancher/rancher/issues/21253

Problem:
When upgrading a project scoped catalog app, the externalId is not constructed properly. Therefore, it results in a version-not-found error. The cause is that URL parsing does not preserve query orders, given that it turns to a map under the hood. It breaks since we introduced 'type' param to the externalId. 

For example, 
`catalog://?catalog=cid&type=projectCatalog&template=grafana&version=0.0.30` 
becomes 
`catalog://?catalog=cid&template=grafana&type=projectCatalog&version=0.0.30`. 
Therefore the externalId filter does not work.

Solution:
Change the url.Encode way to get externalId for a newer version

### DIFF
--- a/cmd/app.go
+++ b/cmd/app.go
@@ -453,14 +453,9 @@ func updateExternalIDVersion(externalID string, version string) (string, error) 
 		return "", err
 	}
 
-	q := u.Query()
-	q.Set("version", version)
-	// Need to unescape here to get the raw string for the externalId filter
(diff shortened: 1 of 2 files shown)


orangedeng:
LGTM

rmweir:
LGTM

dramich:
lgtm

PR #240, opened 2019-07-24
Allow ssh via external ip

dhrp:
This PR implements a change so that you may SSH to the external node IP with the rancher CLI

```
rancher -e node-name
```

It also changes the flag for specifying the user from -l to -u (which seems to make more sense)

All changes are actually from jeremyweber, and I found them through:
https://forums.rancher.com/t/rancher-cli-and-ssh/11107

But since he was not replying and since he never made a PR, and I'd still love to have this contributed to upstream, here is a PR..

### DIFF
--- a/cmd/ssh.go
+++ b/cmd/ssh.go
@@ -20,15 +20,21 @@ import (
 )
 
 const sshDescription = `
-For any nodes created through Rancher using docker-machine, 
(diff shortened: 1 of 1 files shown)


dhrp:
I'm blind. There was a PR for this..

PR #242, opened 2019-07-30
Jweber allow ssh external ip

dhrp:
This PR implements a change so that you may SSH to the external node IP with the rancher CLI

rancher -e node-name

Most changes are actually from jeremyweber, in https://github.com/rancher/cli/pull/185 but that PR has gone stale.

I have also implemented the feedback on that PR (reverted login name back to -l flag). (Also found that one description line was duplicated)

### DIFF
--- a/cmd/ssh.go
+++ b/cmd/ssh.go
@@ -20,15 +20,21 @@ import (
 )
 
 const sshDescription = `
-For any nodes created through Rancher using docker-machine, 
+For any nodes created through Rancher using docker-machine,
 you can SSH into the node. This is not supported for any custom nodes.
 
 Examples:
 	# SSH into a node by ID/name
 	$ rancher ssh nodeFoo
 
+	# SSH into a node by ID/name using the external IP address
+	$ rancher ssh -e nodeFoo
+
 	# SSH into a node by name but specify the username to use
-	$ rancher ssh -l user1 nodeFoo
+	$ rancher ssh -u user1 nodeFoo
+
+	# SSH into a node by specifying user and node using the @ syntax while adding a command to run
+	$ rancher ssh user1@nodeFoo env
 
 	# SSH into a node by specifying user and node using the @ syntax while adding a command to run
 	$ rancher ssh user1@nodeFoo env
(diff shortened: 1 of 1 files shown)


cmd/ssh.go
```diff
@@ -20,59 +20,65 @@ import (
...
+	# SSH into a node by ID/name using the external IP address
+	$ rancher ssh -e nodeFoo
 
-	# SSH into a node by specifying user and node using the @ syntax while adding a command to run
-	$ rancher ssh user1@nodeFoo env
+	# SSH into a node by name but specify the login name to use
+	$ rancher ssh -l login1 nodeFoo
+
+	# SSH into a node by specifying login name and node using the @ syntax while adding a command to run
+	$ rancher ssh login1@nodeFoo env
```
dramich: Update this example to include `--` for example `$ rancher ssh login1@nodeFoo -- netstat -p tcp`. The `--` is needed or flag parsing will drop flags from the command being passed through (`-p`) in this case.

cmd/ssh.go
```diff
@@ -20,59 +20,65 @@ import (
...
-		SkipFlagParsing: true,
+		Name:        "ssh",
+		Usage:       "SSH into a node",
+		Description: sshDescription,
+		Action:      nodeSSH,
+		ArgsUsage:   "[NODE_ID/NODE_NAME]",
+		Flags: []cli.Flag{
+			cli.BoolFlag{
+				Name:  "external,e",
+				Usage: "Use the external ip address",
```
dramich:
```suggestion
				Usage: "Use the external ip address of the node",
```

cmd/ssh.go
```diff
@@ -20,59 +20,65 @@ import (
...
+			},
+			cli.StringFlag{
+				Name:  "login,l",
+				Usage: "The login name",
+			},
+		},
+		SkipFlagParsing: false,
```
dramich: This can be dropped. The default is false.

cmd/ssh.go
```diff
@@ -20,59 +20,65 @@ import (
...
 	}
 }
 
 func nodeSSH(ctx *cli.Context) error {
-	if ctx.NArg() == 0 {
-		return cli.ShowCommandHelp(ctx, "ssh")
-	}
```
dramich: Nit: extra line here

dramich:
The functionality looks good, some small comments.

dramich:
https://github.com/rancher/rancher/issues/13556

PR #243, opened 2019-07-31
#19925 add rancher server command

chaudyg:
As per https://github.com/rancher/rancher/issues/19925, add the following commands
* rancher server current
* rancher server ls
* rancher server switch
* rancher server delete

### DIFF
--- a/cmd/server.go
+++ b/cmd/server.go
@@ -1,0 +1,155 @@
+package cmd
+
+import (
+	"bufio"
+	"fmt"
+	"os"
+	"strings"
+
+	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
+	"github.com/urfave/cli"
+)
+
+type serverData struct {
+	Name string
+	URL  string
+}
+
+func ServerCommand() cli.Command {
+	return cli.Command{
+		Name:  "server",
+		Usage: "Operations for the server",
+		Description: `Switch or view the server currently in focus.
+`,
+		Subcommands: []cli.Command{
+			{
+				Name:   "current",
+				Usage:  "Display the current server",
+				Action: serverCurrent,
+			},
+			{
+				Name:      "delete",
+				Usage:     "Delete a server",
+				ArgsUsage: "[SERVER_NAME]",
+				Description: `
+The server arg is optional, if not passed in a list of available servers will
+be displayed and one can be selected.
+`,
+				Action: serverDelete,
+			},
+			{
+				Name:   "ls",
+				Usage:  "List all servers",
+				Action: serverLs,
+			},
+			{
+				Name:      "switch",
+				Usage:     "Switch to a new server",
+				ArgsUsage: "[SERVER_NAME]",
+				Description: `
+The server arg is optional, if not passed in a list of available servers will
+be displayed and one can be selected.
+`,
+
+				Action: serverSwitch,
+			},
+		},
+	}
+}
+
+func serverCurrent(ctx *cli.Context) error {
+	cf, err := loadConfig(ctx)
+	if err != nil {
+		return err
+	}
+
+	fmt.Printf("%s\n", cf.CurrentServer)
+	return nil
+}
+
+func serverDelete(ctx *cli.Context) error {
+	cf, err := loadConfig(ctx)
+	if err != nil {
+		return err
+	}
+
+	serverName, err := serverSelect(ctx)
+	if err != nil {
+		return err
+	}
+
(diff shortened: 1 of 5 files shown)


cmd/server.go
```diff
@@ -0,0 +1,155 @@
...
+	}
+}
+
+func serverCurrent(ctx *cli.Context) error {
+	cf, err := loadConfig(ctx)
+	if err != nil {
+		return err
+	}
+
+	fmt.Printf("%s\n", cf.CurrentServer)
```
dramich: Give more context with this `Name:xxx URL:xxx`

cmd/server.go
```diff
@@ -0,0 +1,155 @@
...
+}
+
+func serverLs(ctx *cli.Context) error {
+	cf, err := loadConfig(ctx)
+	if err != nil {
+		return err
+	}
+
+	writer := NewTableWriter([][]string{
+		{"NAME", "Name"},
```
dramich: Include current here, similar to running `rancher cluster`

cmd/server.go
```diff
@@ -0,0 +1,155 @@
...
+		return errors.New("Server not found")
+	}
+
+	cf.CurrentServer = serverName
+	cf.Write()
+
+	return nil
+}
+
+func serverSelect(ctx *cli.Context) (string, error) {
```
dramich: This should operate just like `login` does [here](https://github.com/rancher/cli/blob/master/cmd/login.go#L205). Having to type the name isn't intuitive and it's different from how the other selects work.

cmd/server.go
```diff
@@ -0,0 +1,155 @@
...
+	if err != nil {
+		return err
+	}
+
+	_, ok := cf.Servers[serverName]
+	if !ok {
+		return errors.New("Server not found")
+	}
+
+	cf.CurrentServer = serverName
```
dramich: It would be a good idea to check if a current project/cluster is set and if not warn the user `No context set; some commands will not work. Run `rancher context switch` to set one` or similar.

cmd/server.go
```diff
@@ -0,0 +1,155 @@
...
+`,
+		Subcommands: []cli.Command{
+			{
+				Name:   "current",
+				Usage:  "Display the current server",
+				Action: serverCurrent,
+			},
+			{
+				Name:      "delete",
+				Usage:     "Delete a server",
```
dramich: `"Delete a server from the local config"`

dramich:
This is looking good, thanks for doing it.

chaudyg:
@dramich Implemented the changes you suggested. Let me know what you think.

Also, I could not find any test in the repo.  What are the plans regarding testing? Shall I write a few tests as part of this PR and update the drone pipeline to run them?

cmd/server.go
```diff
@@ -0,0 +1,227 @@
...
+
+// serverFromInput displays the list of servers from the local config and
+// prompt the user to select one.
+func serverFromInput(ctx *cli.Context, cf config.Config) (string, error) {
+
+	displayListServers(ctx, cf)
+	fmt.Print("Select a Server:")
+
+	reader := bufio.NewReader(os.Stdin)
+	input, err := reader.ReadString('\n')
```
dramich: Lift [this](https://github.com/rancher/cli/blob/master/cmd/login.go#L228-L253) for use here. This will handle the case that the number is outside the range and inputs that aren't a number.

cmd/server.go
```diff
@@ -0,0 +1,227 @@
...
+	if input == "" {
+
+	}
+
+	i, err := strconv.Atoi(input)
+	if err != nil {
+		return "", err
+	}
+
+	serverNames := getServerNames(cf)
```
dramich: move `serverNames := getServerNames(cf)` to the top then you can use it with the for loop and when you return at the end drop the call to `getServerNames`

dramich:
@chaudyg You can write tests if you want but they are not required. If you add go tests they will get ran in our CI so no other changes are needed.

If you do add tests please use the `github.com/stretchr/testify/assert` library for asserts. [Here](https://github.com/rancher/rancher/blob/master/pkg/httpproxy/proxy_test.go#L143) is an example of some tests.

cmd/server.go
chaudyg: Done

cmd/server.go
chaudyg: Done

rmweir:
lgtm

PR #244, opened 2019-08-05
Pass rancherVersion when fetching templates

dramich:
Problem:
Rancher supports enforcing apps versions but all app templates are shown
even if they are not compatable with the version of rancher running

Solution:
Add the rancherVersion filter when getting templates so only valid app
templates are shown.

This does not stop a user from passing a version that is not considered valid for their Rancher server, the API enforces version checking for all app actions. 

Issue: https://github.com/rancher/rancher/issues/21733

### DIFF
--- a/cmd/app.go
+++ b/cmd/app.go
@@ -573,7 +573,7 @@ func templateShow(ctx *cli.Context) error {
 		return err
 	}
 
-	template, err := c.ManagementClient.Template.ByID(resource.ID)
+	template, err := getFilteredTemplate(ctx, c, resource.ID)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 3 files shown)


cmd/app.go
```diff
@@ -948,6 +960,27 @@ func parseExternalID(e string) (map[string]string, error) {
 	return parsed, nil
 }
 
+func getFilteredTemplate(ctx *cli.Context, c *cliclient.MasterClient, templateID string) (*managementClient.Template, error) {
```
luthermonson: feels weird to call this getFilteredTemplate, you dont allow any filters and just pull and auto set the version. not critical tho

luthermonson:
lgtm

cjellick:
lgtm

PR #246, opened 2019-08-09
Check for cluster export action

dramich:
Problem:
Some clusters can not be exported which can cause a unfriendly error

Solution:
Validate a cluster has the export action before attempting the call

Issue: https://github.com/rancher/rancher/issues/21309


luthermonson:
LGTM

PR #252, opened 2019-09-03
Print actual mcapp name

rmweir:
Problem:
If the user did not
pass a name, then a blank name was printed.

Solution:
Installing print text now uses name from created app. Prior, printed
app name was taken from user passed arguments.

Issue:
https://github.com/rancher/rancher/issues/22527

### DIFF
--- a/cmd/multiclusterapp.go
+++ b/cmd/multiclusterapp.go
@@ -611,9 +611,10 @@ func multiClusterAppRollback(ctx *cli.Context) error {
 }
 
 func multiClusterAppTemplateInstall(ctx *cli.Context) error {
-	if ctx.NArg() >= 2 {
+	if ctx.NArg() > 2 {
 		return cli.ShowSubcommandHelp(ctx)
 	}
+
 	templateName := ctx.Args().First()
 	appName := ctx.Args().Get(1)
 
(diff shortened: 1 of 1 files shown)


cmd/multiclusterapp.go
```diff
@@ -611,9 +611,10 @@ func multiClusterAppRollback(ctx *cli.Context) error {
 }
 
 func multiClusterAppTemplateInstall(ctx *cli.Context) error {
-	if ctx.NArg() >= 2 {
+	if ctx.NArg() >= 3 {
```
luthermonson: maybe > 2?

cmd/multiclusterapp.go
```diff
@@ -722,12 +723,14 @@ func multiClusterAppTemplateInstall(ctx *cli.Context) error {
...
 	app.Timeout = ctx.Int64("helm-timeout")
 
-	_, err = c.ManagementClient.MultiClusterApp.Create(app)
+	app, err = c.ManagementClient.MultiClusterApp.Create(app)
 	if err != nil {
 		return err
 	}
 
-	fmt.Printf("Installing multi-cluster app %q...\n", appName)
+	if app != nil {
```
dramich: Why the nil check? if there is no error there should be an app. Have you seen differently?

dramich:
lgtm

PR #254, opened 2019-10-01
Call yaml.Unmarshal on values/answers file using YAML

mrajashree:
https://github.com/rancher/rancher/issues/20704
We use json.Unmarshal for files using JSON format, and convert the
YAML format files to JSON first and call json.Unmarshal on them as well.
The yaml to json conversion led to unquoted numbers getting converted to
scientific notation. So this commit fixes it by calling yaml.Unmarshal
directly on file with yaml format instead of converting them to json.

### DIFF
--- a/cmd/app.go
+++ b/cmd/app.go
@@ -1078,13 +1078,13 @@ func processAnswers(
 	}
 
 	if ctx.String("values") != "" {
-		if err := getValuesFile(ctx.String("values"), answers); err != nil {
+		if err := parseValuesFile(ctx.String("values"), answers); err != nil {
 			return answers, err
 		}
 	}
 
 	if ctx.String("answers") != "" {
-		err := getAnswersFile(ctx.String("answers"), answers)
+		err := parseAnswersFile(ctx.String("answers"), answers)
 		if err != nil {
 			return answers, err
 		}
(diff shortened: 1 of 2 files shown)


cmd/app.go
```diff
@@ -1131,20 +1124,36 @@ func getAnswersFile(location string, answers map[string]string) error {
...
 
+func parseFile(location string) (map[string]interface{}, error) {
+	bytes, err := ioutil.ReadFile(location)
+	if err != nil {
+		return nil, err
+	}
+
+	values := make(map[string]interface{})
+	if hasPrefix(bytes, []byte("{")) {
+		// this is the check that "readFileReturnJSON" uses to differentiate between JSON and YAML
```
leodotcloud: Can we get rid of the func `readFileReturnJSON` altogether from the repo?

mrajashree: So it's still being used in getRKEConfig, didn't want to change that area with this pr

dramich: You can leave it, it would require a bigger refactor since everything is working with bytes

mrajashree: Based on offline discussion with @dramich and @leodotcloud , we'll keep `readFileReturnJSON` for `getRKEConfig`, since that method expects the output from `readFileReturnJSON` in bytes and calls a couple more methods on it before calling unmarshal. Making all those changes seems out of the scope of this pr, especially since we don't have a proper way of testing because currently creating cluster by passing rke-config only extracts top level keys.

dramich:
lgtm

PR #257, opened 2019-10-30
Corrected issue where rancher app upgrades were not passing through c…

GL-KenDLG:
…onfig from the existing deployment/app See https://github.com/rancher/rancher/issues/23742 for more details


superseb:
This should probably be fixed by https://github.com/rancher/cli/pull/267

PR #266, opened 2020-03-26
Add helm version flag

brendarearden:
**Problem**
Helm version is set at the catalog level, but needs the ability to set it at the catalog level through the cli

**Solution**
Add a flag to the catalog add options to set the helm-version 

**Issue**
https://github.com/rancher/rancher/issues/25262


cmd/catalog.go
```diff
@@ -19,6 +20,9 @@ Example:
 
 	# Add a catalog and specify the branch to use
 	$ rancher catalog add --branch awesomebranch foo https://my.catalog
+
+	# Add a catalog and specify the helm version to use
```
rmweir: We should probably specify the two choices are v2 and v3

cbron: Yea or they might use `helmV3` or `3`. Otherwise lgtm.

rmweir: or try using a version that has a minor or patch

PR #267, opened 2020-03-31
Values yaml

brendarearden:
**Problem**
When a values yaml was being passed through the cli, it would set the answers but when the template was getting edited in the ui, it would update the valuesYaml

**Solution**
Set valuesYaml when value is being used instead of answers

**Issue**
https://github.com/rancher/rancher/issues/23410
https://github.com/rancher/rancher/issues/21083

### DIFF
--- a/cmd/app.go
+++ b/cmd/app.go
@@ -431,19 +431,9 @@ func appUpgrade(ctx *cli.Context) error {
 	if err != nil {
 		return err
 	}
-
-	answers := app.Answers
-	if ctx.String("answers") != "" {
-		answers, err = processAnswers(ctx, c, nil, answers, false)
-		if err != nil {
-			return err
-		}
-	}
-
-	values := app.ValuesYaml
-	if ctx.String("values") != "" {
-		values, err = parseFileToString(ctx.String("values") )
-		answers = nil
+	answers, values, err := getAnswersAndValues(ctx, c, app.Answers, false)
+	if err != nil {
+		return err
 	}
 
 	force := ctx.Bool("force")
@@ -451,7 +441,7 @@ func appUpgrade(ctx *cli.Context) error {
 	au := &projectClient.AppUpgradeConfig{
 		Answers:      answers,
 		ForceUpgrade: force,
-		ValuesYaml: values,
+		ValuesYaml:   values,
 	}
 
 	if resolveTemplatePath(appVersionOrLocalTemplatePath) {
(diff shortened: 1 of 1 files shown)


mrajashree:
@brendarearden it looks like your PR would also fix this bug? I was considering a similar approach for it https://github.com/rancher/rancher/issues/21083#issuecomment-554076368

cmd/app.go
```diff
@@ -433,16 +433,25 @@ func appUpgrade(ctx *cli.Context) error {
...
+		answers, err = processAnswers(ctx, c, nil, answers, false)
+		if err != nil {
+			return err
+		}
+	}
+
+	values := app.ValuesYaml
+	if ctx.String("values") != "" {
+		values, err = parseFileToString(ctx.String("values"))
+		answers = nil
```
mrajashree, line 446: It looks like the helm controller accepts both, answers and values and appends the passed answers to values as per https://github.com/rancher/rancher/blob/master/pkg/controllers/user/helm/common/common.go#L216-L240
So we don't need to set answers to nil here?

cmd/app.go
```diff
@@ -433,16 +433,25 @@ func appUpgrade(ctx *cli.Context) error {
...
+	if ctx.String("answers") != "" {
+		answers, err = processAnswers(ctx, c, nil, answers, false)
+		if err != nil {
+			return err
+		}
+	}
+
+	values := app.ValuesYaml
+	if ctx.String("values") != "" {
+		values, err = parseFileToString(ctx.String("values"))
```
mrajashree, line 445: is it possible to extract this logic in a separate function and call that here and in templateInstall as it seems common to both?

cmd/app.go
brendarearden: based on how the ui works - we should return answers or values, not both

brendarearden:
@mrajashree I should fix that bug you linked

cmd/app.go
```diff
@@ -1093,17 +1093,9 @@ func processAnswers(
 		answers = make(map[string]string)
 	}
 
-	if ctx.String("values") != "" {
-		if err := parseValuesFile(ctx.String("values"), answers); err != nil {
-			return answers, err
-		}
-	}
-
-	if ctx.String("answers") != "" {
```
cbron, line 1102: Why remove the check ?

cbron: nvm

cmd/app.go
```diff
@@ -1308,3 +1299,24 @@ func checkShowSubquestionIf(q managementClient.Question, answers map[string]stri
 	}
 	return false
 }
+
+// return answers and empty values or values and empty answers to match UI flow
```
cbron, line 1303: Couldn't both be returned here ? Or would both not be passed in

brendarearden: I will make the comment more clear, but essentially if you pass in answers OR values, the other one should be empty, if you pass in BOTH in the cli command - both should return values

cbron: Thats more clear, although code looks like if you passed both it would return both, thats why comment confused me.

PR #268, opened 2020-04-01
Use default answers when setting no-prompt flag

brendarearden:
**Problem**
On install, answers were only being generated if prompted for answers

**Solution**
Build answers with default responses when --no-prompt is true

**Issue**
https://github.com/rancher/rancher/issues/25514

### DIFF
--- a/cmd/app.go
+++ b/cmd/app.go
@@ -1119,6 +1119,8 @@ func processAnswers(
 		if err != nil {
 			return answers, err
 		}
+	} else {
+		answers = fillInDefaultAnswers(tv, answers)
 	}
 
 	return answers, nil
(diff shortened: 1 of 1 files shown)


cmd/app.go
```diff
@@ -1280,6 +1282,23 @@ func askSubQuestion(q managementClient.SubQuestion) string {
 	return answer
 }
 
+func getAllDefaultAnswers(tv *managementClient.TemplateVersion, answers map[string]string) map[string]string {
```
cbron, line 1285: nit: this name sounds like you are setting all answers to default, when in reality you are filling in missing answers with the defaults. Maybe `fillInDefaultAnswers` ?

cmd/app.go
```diff
@@ -1280,6 +1282,23 @@ func askSubQuestion(q managementClient.SubQuestion) string {
...
+	for _, question := range tv.Questions {
+		if _, ok := answers[question.Variable]; !ok && checkShowIf(question.ShowIf, answers) {
+			answers[question.Variable] = question.Default
+			if checkShowSubquestionIf(question, answers) {
+				for _, subQuestion := range question.Subquestions {
+					// only ask the question if there is not an answer and it passes the ShowIf check
```
daxmc99, line 1291: Hmm, maybe this comment should be rewritten? We should only set the subquestion if it passes the showIf check? Asking the question sounds interactive and it appears this prompt should not be interactive?

PR #273, opened 2020-04-16
Set answers to question defaults

brendarearden:
**Problem**
On install, answers were only being generated if prompted for answers or when being passed in

**Solution**
Build answers with default question responses when --no-prompt is true

**Original PR** 
https://github.com/rancher/cli/pull/268

**How it was previously working**
when the no-prompt flag is used, it will bypass being prompted for questions

**How this PR is working** 
When the no-prompt flag is used, it will fill in the valuesYaml with the default answers to the questions without being prompted to answer the questions. By setting the defaults in the valuesYaml, it follows more closely the hierarchy of answers / values / values yaml for helm charts

**Issue**
rancher/rancher#25514

### DIFF
--- a/cmd/app.go
+++ b/cmd/app.go
@@ -431,7 +431,10 @@ func appUpgrade(ctx *cli.Context) error {
 	if err != nil {
 		return err
 	}
-	answers, values, err := getAnswersAndValues(ctx, c, app.Answers, false)
+
+	answers := app.Answers
+	values := app.ValuesYaml
+	answers, values, err = processFileInputs(ctx, c, nil, answers, values, false)
 	if err != nil {
 		return err
 	}
@@ -628,8 +631,7 @@ func templateInstall(ctx *cli.Context) error {
 		if err != nil {
 			return err
 		}
-
-		answers, values, err := getAnswersAndValues(ctx, c, app.Answers, false)
+		answers, values, err := processFileInputs(ctx, c, nil, nil, "", false)
 		if err != nil {
 			return err
 		}
@@ -682,10 +684,11 @@ func templateInstall(ctx *cli.Context) error {
 		}
 
 		interactive := !ctx.Bool("no-prompt")
-		answers, values, err := getAnswersAndValues(ctx, c, app.Answers, interactive)
+		answers, values, err := processFileInputs(ctx, c, templateVersion, nil, "", interactive)
 		if err != nil {
 			return err
 		}
+
 		namespace := ctx.String("namespace")
 		if namespace == "" {
 			namespace = template.Name + "-" + RandomLetters(5)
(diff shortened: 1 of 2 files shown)


cmd/app.go
```diff
@@ -1104,12 +1103,18 @@ func processAnswers(
...
 		}
 	}
-
-	if interactive {
-		// answers to questions will be added to map
-		err := askQuestions(tv, answers)
-		if err != nil {
-			return answers, err
+	// checks if answers or values were passed in before prompting with questions or setting question defaults
+	if ctx.String("answers") == "" && ctx.String("values") == "" {
```
rmweir, line 1107: Question: do we want to skip this just because some answers are provided? Is it possible that a values in values.yaml is not covered by questions.yaml?

cmd/app.go
```diff
@@ -1221,6 +1291,40 @@ func askQuestions(tv *managementClient.TemplateVersion, answers map[string]strin
 	}
 }
 
+func askQuestionsNotFoundInValues(tv *managementClient.TemplateVersion, tempAnswers map[string]string) (map[string]string, error) {
```
cbron, line 1294: Is the only difference from original method that the conditional checks are against tempAnswers but the new answers map is returned ? Would it be have same outcome to just add a boolean flag to that method and then determine at the return point what data to return ?

brendarearden: yes - I combined the methods per this suggestion

cmd/app.go
```diff
@@ -1105,15 +1140,36 @@ func processAnswers(
 		}
 	}
 
+	// answers to questions will be added to map
 	if interactive {
-		// answers to questions will be added to map
-		err := askQuestions(tv, answers)
-		if err != nil {
-			return answers, err
+		if ctx.String("values") != "" {
```
rmweir, line 1145: This looks like duplicate logic of the block starting at L1101 to me, are we not askQuestions twice?

brendarearden: removed duplicate logic

cmd/app.go
```diff
@@ -1080,40 +1082,91 @@ func createNamespace(c *cliclient.MasterClient, n string) error {
...
-	err := parseAnswersFile(ctx.String("answers"), answers)
-	if err != nil {
-		return answers, err
+	if ctx.String("answers") != "" {
+		err := parseAnswersFile(ctx.String("answers"), answers)
+		if err != nil {
+			return answers, values, err
+		}
 	}
+	fmt.Println("answersAfterAnswers: ", answers)
```
cbron, line 1121: Debug print ?

cmd/app.go
```diff
@@ -1080,40 +1082,91 @@ func createNamespace(c *cliclient.MasterClient, n string) error {
...
+			if err != nil {
+				return answers, values, err
+			}
+		}
+	} else {
+		if ctx.String("values") != "" {
+			tempAnswers, err = fillInDefaultAnswers(tv, tempAnswers, false)
+			for k, v := range answers {
+				tempAnswers[k] = v
+			}
```
rmweir, line 1152: isn't this happening at L1165?

brendarearden: yes - I moved it to 1165 but failed to remove it here. good catch! will remove duplicate code

cmd/app.go
cbron: one above too

cmd/app.go
```diff
@@ -1080,40 +1082,91 @@ func createNamespace(c *cliclient.MasterClient, n string) error {
...
 		}
 	}
-
-	return answers, nil
+	// ensures that if both answers and values exists, they are both assigned to answers
+	for k, v := range answers {
+		tempAnswers[k] = v
+	}
+	answers = tempAnswers
```
cbron, line 1168: just return tempAnswers

cmd/app.go
```diff
@@ -1080,6 +1083,92 @@ func createNamespace(c *cliclient.MasterClient, n string) error {
...
+	c *cliclient.MasterClient,
+	tv *managementClient.TemplateVersion,
+	answers map[string]string,
+	values string,
+	interactive bool) (map[string]string, string, error) {
+	var err error
+	if ctx.String("answers") != "" && ctx.String("values") != "" {
+		return nil, "", fmt.Errorf("cannot process an answers file and a values file")
+	}
+	if hasAnswers(ctx.String("answers"), answers) || !hasAnswers(ctx.String("answers"), answers) && !hasValues(ctx.String("values"), values) {
```
rmweir, line 1097: you can remove the second `!hasAnswers(ctx.String("answers"), answers)` as the condition won't be evaluated unless that is the case

cmd/app.go
```diff
@@ -1080,6 +1083,92 @@ func createNamespace(c *cliclient.MasterClient, n string) error {
...
+	}
+	return false
+}
+
+// creates a map of the values file and fills in missing answers with defaults
+func processValues(ctx *cli.Context, tv *managementClient.TemplateVersion, values string) (string, error) {
+	answers := make(map[string]string)
+	if values != "" {
+		//parse values into map to ensure previous values are considered on update
+		defaultsMap, err := createValuesMap([]byte(values))
```
rmweir, line 1133: I think we shouldn't call this defaults map if we are starting off by copy non-default values to it

cmd/app.go
```diff
@@ -1080,6 +1083,92 @@ func createNamespace(c *cliclient.MasterClient, n string) error {
...
+
+func hasValues(filePathString string, valuesString string) bool {
+	if filePathString != "" || valuesString != "" {
+		return true
+	}
+	return false
+}
+
+// creates a map of the values file and fills in missing answers with defaults
+func processValues(ctx *cli.Context, tv *managementClient.TemplateVersion, values string) (string, error) {
```
rmweir, line 1129: can we call values something like existingValues? I didn't immediately understand how we were using this or how it was different from the values that could be passed.

cmd/app.go
```diff
@@ -1080,6 +1083,92 @@ func createNamespace(c *cliclient.MasterClient, n string) error {
...
+	// change map back into string to be consistent with ui
+	values, err = parseMapToYamlString(answers)
+	if err != nil {
+		return values, err
+	}
+	return values, nil
+}
+
+// returns map with default answers
+func createDefaultMap(tv *managementClient.TemplateVersion, answers map[string]string) error {
```
rmweir, line 1161: Do we need this function? it looks like it just calls fillInDefaults. If we were to keep it I do think it should be renamed to something like applyDefaults since its not actually creating anything.

cmd/app.go
```diff
@@ -1080,6 +1083,92 @@ func createNamespace(c *cliclient.MasterClient, n string) error {
...
+	if values != "" {
+		//parse values into map to ensure previous values are considered on update
+		defaultsMap, err := createValuesMap([]byte(values))
+		if err != nil {
+			return values, err
+		}
+		valuesToAnswers(defaultsMap, answers)
+	}
+	// add default values if answers missing from map
+	err := createDefaultMap(tv, answers)
```
rmweir, line 1140: Can probably replace this with fillInDefaults, see other comment at L1161

cmd/app.go
```diff
@@ -1080,6 +1083,80 @@ func createNamespace(c *cliclient.MasterClient, n string) error {
...
+	}
+	if hasAnswers(ctx.String("answers"), answers) || !hasValues(ctx.String("values"), values) {
+		values = ""
+		answers, err = processAnswers(ctx, c, tv, answers, interactive)
+		if err != nil {
+			return answers, values, err
+		}
+	}
+	if hasValues(ctx.String("values"), values) {
+		answers = nil
```
dramich, line 1105: What will happen here if a user does have both set on the app and we go to do an upgrade?

cmd/app.go
```diff
@@ -1080,6 +1083,80 @@ func createNamespace(c *cliclient.MasterClient, n string) error {
...
+}
+
+func hasValues(filePathString string, valuesString string) bool {
+	if filePathString != "" || valuesString != "" {
+		return true
+	}
+	return false
+}
+
+// creates a map of the values file and fills in missing answers with defaults
```
dramich, line 1128: Go comments for functions always start with the function name

cmd/app.go
brendarearden: it will hit the hasAnswers block first so it will keep and update the answers and empty the values, unless they are passing in a values file.

dramich: If both are set I don't think we should blank out anything.

brendarearden: added support for both values to be kept unless new files are passed in

PR #274, opened 2020-04-23
Revert "Values yaml"

brendarearden:
Reverts rancher/cli#267


brendarearden:
Did not need to be done

PR #281, opened 2020-07-08
Fix values error by combining values with answers

yogeshmundada:
Issue: https://github.com/rancher/rancher/issues/27841

Cause: values are converted into string. Thus bool values get converted to "false"/"true". The ternary sprig operator expects the bool values instead of a string. However, in rancher/types, values are of type string. Answers get marshalled/unmarshalled correctly though. So to fix it at the moment combining new values with the answers.

Without the fix, the output from rancher/rancher shows:

```
yogesh@Yogeshs-MacBook-Pro:~/go/src/rancher/cli$ ./cli apps install --values wp-values.yaml  wordpress --version 9.0.3 --no-prompt
run "app show-notes a-vszh9" to view app notes once app is ready
yogesh@Yogeshs-MacBook-Pro:~/go/src/rancher/cli$
```

rancher/rancher logs:
```
2020/07/08 13:57:42 [DEBUG] [helm-controller] Updating app a-vszh9
2020/07/08 13:57:42 [DEBUG] Helm preparing catalog cache [library] for update
2020/07/08 13:57:42 [DEBUG] Helm updated git-helm catalog [https://git.rancher.io/charts:master]
2020/07/08 13:57:42 [ERROR] AppController p-t4gdv/a-vszh9 [helm-controller] failed with : wait helm template failed. Error: render error in "wordpress/templates/deployment.yaml": template: wordpress/templates/deployment.yaml:58:50: executing "wordpress/templates/deployment.yaml" at <.Values.allowEmptyPassword>: wrong type for value; expected bool; got string
: exit status 1
```

With the fix the error is gone.

### DIFF
--- a/cmd/app.go
+++ b/cmd/app.go
@@ -438,7 +438,14 @@ func appUpgrade(ctx *cli.Context) error {
 	if err != nil {
 		return err
 	}
-	values, err = processValueUpgrades(ctx, values)
+
+	var unmarshalledValues map[string]interface{}
+	err = yaml.Unmarshal([]byte(app.ValuesYaml), &unmarshalledValues)
+	if err != nil {
+		return err
+	}
+	valuesToAnswers(unmarshalledValues, answers)
+	values, err = processValueUpgrades(ctx, answers)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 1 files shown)


yogeshmundada:
I am abandoning this approach and have generated another PR.

https://github.com/rancher/cli/pull/282

PR #282, opened 2020-07-10
Fix the values processing on cli

yogeshmundada:
Problem: rancher/rancher#27841

Solution: Instead of parsing the yaml as a map[string]string that would lose type information about bool variables, now parsing it as map[string]interface{}. When flattening the structure, a separate case for bool type is created so that other map members are converted to string except bool.

### DIFF
--- a/cmd/app.go
+++ b/cmd/app.go
@@ -1130,8 +1130,8 @@ func processValueUpgrades(ctx *cli.Context, existingValues string) (string, erro
 }
 
 // processValues creates a map of the values file
-func processValues(ctx *cli.Context, existingValues string) (map[string]string, error) {
-	answers := make(map[string]string)
+func processValues(ctx *cli.Context, existingValues string) (map[string]interface{}, error) {
+	answers := make(map[string]interface{})
 	if existingValues != "" {
 		//parse values into map to ensure previous values are considered on update
 		valuesMap, err := createValuesMap([]byte(existingValues))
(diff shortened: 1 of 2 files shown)


yogeshmundada:
With this fix, most of the issues will be fixed.
There is one discrepancy remaining. If the UI is edited for something like "abc: yes", then it will treat it as a string.
However, when installed from the CLI, it would be converted to true. A related issue can be fond here:
https://github.com/uber-go/config/issues/74

However, without addition of special tags around these specific values, there does not seem to be a straightforward way to fix this.

yogeshmundada:
> Using this fix will introduce a new bug to the parsing which changes values. This needs a different approach to fully fix the issue.

the "newly introduced bug" is also present in v2.3.2 (except in the answers sections since the past code put everything from --values into answers). The reason behind that is we make a library call yaml.Unmarshal() that already converts an unquoted
```
yes
```
to a boolean value. A quoted one would be retained as a string.

From my understanding of your concern, we don't want incorrect values to be parsed by helm when it upgrades an app.
However, the code to do so in helm is at:
https://github.com/rancher/helm/blob/master/pkg/chartutil/values.go#L119

If we unmarshal the values, we see that it would still result in the same conversion (yes -> true) as below:
https://play.golang.org/p/TZrDkoZqpUm

Please let me know if there is something I am missing. I agree that if someone downloads valuesYAML that is editing in UI and the one uploaded via CLI, and executes a diff on them, then that would not match. However, apart from that, helm should not have any problem loading those values as the playground and gh links above show.

PR #283, opened 2020-07-23
kubeconfig tokens authentication

kinarashah:


cmd/kubectl_token.go
```diff
@@ -0,0 +1,546 @@
...
+	authProvider string
+	caCerts      string
+	skipVerify   bool
+}
+
+const (
+	kubeConfigCache = "/.cache/token"
+	cachedFileExt   = ".json"
+	authProviderURL = "%s/v3-public/authProviders"
+	authTokenURL    = "%sv3-public/authtokens/%s"
```
prachidamle, line 48: Should the url be "authTokens" camelCase just like "authProviders"? or should it be non-camelCase - depends on what the pattern is so far for CLI, but needs to be consistent

cmd/kubectl_token.go
```diff
@@ -0,0 +1,546 @@
...
+			cli.StringFlag{
+				Name:  "user",
+				Usage: "user-id",
+			},
+			cli.StringFlag{
+				Name:  "cluster",
+				Usage: "cluster-id",
+			},
+			cli.StringFlag{
+				Name:  "auth-provider",
```
prachidamle, line 79: should this default to "local"

cmd/kubectl_token.go
```diff
@@ -0,0 +1,546 @@
...
+
+func CredentialCommand() cli.Command {
+	return cli.Command{
+		Name:        "token",
+		Usage:       "Authenticate and generate new kubeconfig token",
+		Description: "todo",
+		Action:      runCredential,
+		Flags: []cli.Flag{
+			cli.StringFlag{
+				Name:  "server",
```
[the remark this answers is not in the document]
prachidamle: does the cli already not ask for this?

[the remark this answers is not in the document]
kinarashah: yep, it's the same. 
cli asks for it, but since this is invoked on `kubectl` calls, it's not necessary that all users would have configured their cli context before that so server gets passed from the kubeconfig generate itself

cmd/kubectl_token.go
kinarashah: we always get `v3-public/authproviders` from the server url to know which providers have been configured -- so not setting default

prachidamle: ok, got it

cmd/kubectl_token.go
```diff
@@ -0,0 +1,594 @@
...
+	}
+
+	responseType := "kubeconfig"
+	if input.clusterID != "" {
+		responseType = fmt.Sprintf("%s_%s", responseType, input.clusterID)
+	}
+
+	tokenURL := fmt.Sprintf(authTokenURL, input.server, id)
+
+	req, err := http.NewRequest(http.MethodGet, tokenURL, bytes.NewBuffer(nil))
```
cbron, line 303: this is supposed to be get and not post ?

cmd/kubectl_token.go
```diff
@@ -0,0 +1,594 @@
...
+	tr := &http.Transport{
+		TLSClientConfig: tlsConfig,
+	}
+
+	client := &http.Client{Transport: tr, Timeout: 300 * time.Second}
+
+	loginRequest := fmt.Sprintf("%s/login?requestId=%s&publicKey=%s&responseType=%s",
+		input.server, id, encodedKey, responseType)
+
+	//openBrowser(input.server)
```
cbron, line 320: Whats this ?

cmd/kubectl_token.go
```diff
@@ -0,0 +1,594 @@
...
+	return response, nil
+}
+
+func customPrompt(field string, show bool) (result string, err error) {
+	fmt.Fprintf(os.Stderr, "Enter %s: ", field)
+	if show {
+		_, err = fmt.Fscan(os.Stdin, &result)
+	} else {
+		var data []byte
+		data, err = terminal.ReadPassword(int(os.Stdin.Fd()))
```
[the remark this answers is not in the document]
kinarashah: we use ReadPassword only for password or secret fields, for authProviders it'd go into the if part https://github.com/rancher/cli/pull/283/files/2af3258cda5f42f0547d354626567aea74adfbeb#diff-5c4870f7fb31cd1a80a523579a64cd37R513

cmd/kubectl_token.go
kinarashah: sorry, it's leftover from my old code, removed

cmd/kubectl_token.go
kinarashah: yep, this is polling `v3-public/authtokens` to read the encrypted token after successful saml login

cmd/kubectl_token.go
```diff
@@ -0,0 +1,594 @@
...
+	if input.clusterID != "" {
+		responseType = fmt.Sprintf("%s_%s", responseType, input.clusterID)
+	}
+
+	body := fmt.Sprintf(`{"responseType":"%s", "username":"%s", "password":"%s"}`, responseType, username, password)
+
+	url := fmt.Sprintf("%sv3-public/%ss/%s?action=login", input.server, input.authProvider,
+		strings.ToLower(strings.Replace(input.authProvider, "Provider", "", 1)))
+
+	response, err := request(http.MethodPost, url, bytes.NewBufferString(body))
```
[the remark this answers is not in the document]
kinarashah: makes sense, renamed it to basicAuth() vs the else part is samlAuth()

cmd/kubectl_token.go
```diff
@@ -0,0 +1,594 @@
...
+	server := ctx.String("server")
+	if server == "" {
+		return fmt.Errorf("name of rancher server is required")
+	}
+	url, err := url2.Parse(server)
+	if err != nil {
+		return err
+	}
+	if url.Scheme == "" {
+		server = fmt.Sprintf("https://%s/", server)
```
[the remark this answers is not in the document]
kinarashah: yep, http's parseURL cleans it up.. other actions require the `/` and it seemed cleaner this way than `%sv3-public` in const

PR #284, opened 2020-08-17
add logic to delete cached credential if it's invalid

kinarashah:
If credential is invalid but is present in the cached directory (token's deleted from rancher), exec plugin doesn't return a proper error msg but just that `server has asked for credentials`. 

Added logic with a statement `use 'rancher token --delete tokenName' if there's an error` so users can run it to delete the cached credential, allowing them to login next time. 

Also updated keycloak provider name.

### DIFF
--- a/cmd/kubectl_token.go
+++ b/cmd/kubectl_token.go
@@ -27,6 +27,20 @@ import (
 	"golang.org/x/crypto/ssh/terminal"
 )
 
+const deleteCommandUsage = `
+Delete cached token used for kubectl login at ${PWD}/.cache/token
+
+Example:
+	# Delete a cached credential 
+	$ rancher token delete cluster1_c-1234
+
+	# Delete multiple cached credentials
+	$ rancher token delete cluster1_c-1234 cluster2_c-2345
+
+	# Delete all credentials 
+	$ rancher token delete all
+`
+
 type LoginInput struct {
 	server       string
 	userID       string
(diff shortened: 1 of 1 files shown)


cmd/kubectl_token.go
```diff
@@ -146,6 +156,27 @@ func runCredential(ctx *cli.Context) error {
 	return json.NewEncoder(os.Stdout).Encode(newCred)
 }
 
+func deleteCachedCredential(ctx *cli.Context) error {
+	dir, err := os.Getwd()
+	if err != nil {
+		return err
+	}
+	cacheDir := filepath.Join(dir, kubeConfigCache)
+	if len(ctx.Args()) == 0 {
```
cbron, line 165: Removing all tokens when not specifying a arg seems heavy. We either need to document this in the param command or force them to set an arg like `all` to get this behavior.

cmd/kubectl_token.go
```diff
@@ -372,7 +414,8 @@ func samlAuth(input *LoginInput, tlsConfig *tls.Config) (managementClient.Token,
 			client = &http.Client{Transport: tr, Timeout: 150 * time.Second}
 			res, err = client.Do(req)
 			if err != nil {
-				return token, err
+				// log error and use the token if login succeeds
+				customPrint(fmt.Sprintf("DeleteToken: %v", err))
```
cbron, line 418: Is this right ? Seems kinda weird message. Looks like you are telling user to do something ?

kinarashah: Sorry it was to be fmt.Errorf...all other msgs are fmt.Errors with what action on token it's failing, and err is the msg

cmd/kubectl_token.go
kinarashah: Yep 💯. I moved it to a subcommand, added `all` to remove all tokens and usage for it

PR #286, opened 2020-10-14
Fix list of maps on chart install w/ user provided values yaml

PennyScissors:
**Problem:**
The Rancher CLI converts user provided values into answers format before marshaling. This results in the API not being able to process lists of maps correctly, and also diverges from how we do it for the UI (Cluster Manager). The chart installation doesn't fail, but the values provided by the user aren't applied.

**Solution:**
Remove the logic that converts values into answers to match the UI marshaling results.

**Test:**
The following was tested and works as intended
- Chart installation passing a values.yaml in the CLI
- Chart upgrading using the CLI, including verifying the previous version's values (user provided values included) were used in the upgraded version.
- Wordpress test case in this [issue](https://github.com/rancher/rancher/issues/27841)
- Test passing in an answers.yaml still works

**Related issues:**
https://github.com/rancher/rancher/issues/28000
https://github.com/rancher/rancher/issues/27841


brendarearden (approved):
looks good!

dramich (approved):

PR #287, opened 2020-10-27
add optional path to cache kubectl token

tjcain:
It would be helpful to have an option to specify a path to the cached kubectl token to avoid changing directories requiring a fresh login.

### DIFF
--- a/cmd/kubectl_token.go
+++ b/cmd/kubectl_token.go
@@ -17,6 +17,7 @@ import (
 	url2 "net/url"
 	"os"
 	"os/signal"
+	"path"
 	"path/filepath"
 	"strings"
 	"time"
@@ -109,6 +110,11 @@ func CredentialCommand() cli.Command {
 				Name:  "skip-verify",
 				Usage: "Skip verification of the CACerts presented by the Server",
 			},
+			cli.StringFlag{
+				Name:   "cache-dir",
+				Usage:  "The absolute path for the kubectl-token .cache directory",
+				EnvVar: "RANCHER_CACHE_DIR",
+			},
 		},
 		Subcommands: []cli.Command{
 			cli.Command{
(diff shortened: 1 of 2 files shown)


cmd/kubectl_token.go
```diff
@@ -109,6 +110,10 @@ func CredentialCommand() cli.Command {
 				Name:  "skip-verify",
 				Usage: "Skip verification of the CACerts presented by the Server",
 			},
+			cli.StringFlag{
+				Name:  "cache-dir",
+				Usage: "The absolute path to place or find a .cache directory containing the kubectl token (also via RANCHER_CACHE_DIR)",
```
dramich, line 115: Add the `EnvVar` field here to specify the envvar and the cli will pull from there automatically
```suggestion
				Usage: "The absolute path for the kubectl-token .cache directory"
```

cmd/kubectl_token.go
```diff
@@ -521,6 +528,23 @@ func generateKey() (string, error) {
 	return string(token), nil
 }
 
+func getCacheDir(ctx *cli.Context) (string, error) {
+	var p string
+	if s := ctx.String("cache-dir"); s != "" {
+		p = s
+	} else if s := os.Getenv("RANCHER_CACHE_DIR"); s != "" {
```
dramich, line 535: Using the `EnvVar` will cover this so it can be dropped

cmd/kubectl_token.go
```diff
@@ -521,6 +528,23 @@ func generateKey() (string, error) {
...
+		p = s
+	} else {
+		return os.Getwd()
+	}
+
+	if !path.IsAbs(p) {
+		return "", fmt.Errorf("cache dir require an absolute path")
```
dramich, line 542:
```suggestion
		return "", fmt.Errorf("cache dir requires an absolute path")
```

cmd/kubectl_token.go
```diff
@@ -164,7 +174,11 @@ func runCredential(ctx *cli.Context) error {
 		return err
 	}
 
-	if err := cacheCredential(newCred, fmt.Sprintf("%s_%s", userID, clusterID)); err != nil {
+	if err != nil {
+		return err
+	}
```
dramich, line 179: Looks like an extra error check

dramich (changes requested):

tjcain:
Thanks for the review, I have made the suggested changes.

dramich (approved):

PR #291, opened 2020-11-12
add global config flag and store token with config

tjcain:
kubeconfig token will now be stored with the rest
of the rancher config rather than in the .cache
directory. The location of the config directory
can now be configured.

### DIFF
--- a/cmd/kubectl_token.go
+++ b/cmd/kubectl_token.go
@@ -17,10 +17,10 @@ import (
 	url2 "net/url"
 	"os"
 	"os/signal"
-	"path/filepath"
 	"strings"
 	"time"
 
+	"github.com/rancher/cli/config"
 	"github.com/rancher/norman/types/convert"
 	managementClient "github.com/rancher/types/client/management/v3"
 	"github.com/urfave/cli"
@@ -51,8 +51,6 @@ type LoginInput struct {
 }
 
 const (
-	kubeConfigCache = "/.cache/token"
-	cachedFileExt   = ".json"
 	authProviderURL = "%s/v3-public/authProviders"
 	authTokenURL    = "%s/v3-public/authTokens/%s"
 )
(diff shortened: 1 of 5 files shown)


cmd/kubectl_token.go
```diff
@@ -175,76 +173,107 @@ func deleteCachedCredential(ctx *cli.Context) error {
...
+	if len(sc.KubeCredentials) == 0 {
+		customPrint(fmt.Sprintf("there are no cached tokens in [%s]", dir))
+		return nil
+	}
+
 	if ctx.Args().First() == "all" {
-		customPrint(fmt.Sprintf("removing cached tokens [%s]", cacheDir))
-		return os.RemoveAll(cacheDir)
+		customPrint(fmt.Sprintf("removing cached tokens [%s]", dir))
+		sc.KubeCredentials = make(map[string]*config.ExecCredential)
```
kinarashah, line 192: when does `sc` get written to config file?

cmd/kubectl_token.go
```diff
@@ -175,76 +173,107 @@ func deleteCachedCredential(ctx *cli.Context) error {
...
 	}
-	cachePath := filepath.Join(dir, kubeConfigCache, fmt.Sprintf("%s%s", key, cachedFileExt))
-	f, err := os.Open(cachePath)
-	if err != nil {
-		if os.IsNotExist(err) {
-			return nil, nil
-		}
 
-		return nil, err
+	return cred, nil
```
kinarashah, line 209: we shouldn't return the credential if it's expired, are we checking this elsewhere? Earlier, it was a part of loadCachedCredential https://github.com/rancher/cli/pull/291/files#diff-06991da8e038887c429ccab4b06868704919c15ade73fca3cd9c4fc8997f7138L217-L219

cmd/kubectl_token.go
tjcain: I missed this, many apologies.

cmd/kubectl_token.go
tjcain: I missed this, many apologies.

kinarashah (approved):
LGTM

dramich (approved):

dramich:
@tjcain Thanks for the work on this, it looks good

PR #292, opened 2020-11-18
Cache kubeconfig used for kubectl commands

rmweir:
Cache and reuse kubeconfig


cmd/common.go
```diff
@@ -139,6 +139,18 @@ func listRoleTemplateBindings(ctx *cli.Context, b []RoleTemplateBinding) error {
...
+// serverCurrent command to display the name of the current server in the local config
+func serverCurrent(ctx *cli.Context) (string, string, error) {
+	cf, err := loadConfig(ctx)
+	if err != nil {
+		return "", "", err
+	}
+
+	serverName := cf.CurrentServer
+	URL := cf.Servers[serverName].URL
+	return serverName, URL, nil
```
nickgerace, line 151: nit:
```go
return serverName, cf.Servers[serverName].URL, nil
```

cmd/kubectl.go
```diff
@@ -5,8 +5,19 @@ import (
...
+	"github.com/sirupsen/logrus"
 	"github.com/urfave/cli"
+	"gopkg.in/yaml.v2"
+
+	"github.com/rancher/norman/clientbase"
+	client "github.com/rancher/types/client/management/v3"
+)
+
+const (
+	kubeconfigpath = ".rancher/.cache/kubeconfig"
```
nickgerace, line 20: Is a relative path always going to work?

cmd/server.go
```diff
@@ -64,15 +64,13 @@ be displayed and one can be selected.
...
+func printServerCurrent(ctx *cli.Context) error {
+	serverName, url, err := serverCurrent(ctx)
 	if err != nil {
 		return err
 	}
 
-	serverName := cf.CurrentServer
-	URL := cf.Servers[serverName].URL
-	fmt.Printf("Name: %s URL: %s\n", serverName, URL)
+	fmt.Printf("Name: %s URL: %s\n", serverName, url)
```
nickgerace, line 73: nit: Should we add a second `\n` character before URL or some other form of separation?

cmd/kubectl.go
```diff
@@ -37,34 +48,66 @@ func runKubectl(ctx *cli.Context) error {
...
 
-	_, err = tmpfile.Write([]byte(config.Config))
+	currentUser := t.UserID
+	currentCluster := currentRancherServer.FocusedCluster()
+	filename := fmt.Sprintf("%s-%s-kubeconfig", currentUser, currentCluster)
+	home, err := os.UserHomeDir()
 	if err != nil {
 		return err
 	}
+	kubeconfigFilePath := fmt.Sprintf("%s/%s/%s", home, kubeconfigpath, filename)
```
nickgerace, line 74: nit: Using `path.Join` or something similar might be better here.
https://golang.org/pkg/path/#Join

nickgerace (approved):
Some small nits, but otherwise, LGTM

cmd/common.go
```diff
@@ -139,6 +139,18 @@ func listRoleTemplateBindings(ctx *cli.Context, b []RoleTemplateBinding) error {
 	return writer.Err()
 }
 
+// serverCurrent command to display the name of the current server in the local config
+func serverCurrent(ctx *cli.Context) (string, string, error) {
```
aiyengar2, line 143: nit: I think this should be renamed to
```suggestion
func getCurrentServerFromConfig(ctx *cli.Context) (string, string, error) {
```

`getCurrentServer` or `currentServer` would be okay as well.

cmd/kubectl.go
```diff
@@ -37,34 +48,66 @@ func runKubectl(ctx *cli.Context) error {
...
+	t, err := c.ManagementClient.Token.ByID(currentToken)
 	if err != nil {
 		return err
 	}
-	defer os.Remove(tmpfile.Name())
 
-	_, err = tmpfile.Write([]byte(config.Config))
+	currentUser := t.UserID
+	currentCluster := currentRancherServer.FocusedCluster()
+	filename := fmt.Sprintf("%s-%s-kubeconfig", currentUser, currentCluster)
```
aiyengar2, line 69: Should this format be a constant at the top of the file?

cmd/kubectl.go
```diff
@@ -37,34 +48,66 @@ func runKubectl(ctx *cli.Context) error {
...
+		}
+		kubeconfig = []byte(config.Config)
+		if err := writeKubeconfig(kubeconfigFilePath, kubeconfig); err != nil {
+			return err
+		}
+	}
+
 	cmd := exec.Command(path, ctx.Args()...)
-	cmd.Env = append(os.Environ(), "KUBECONFIG="+tmpfile.Name())
+	cmd.Env = append(os.Environ(), "KUBECONFIG="+kubeconfigFilePath)
```
aiyengar2, line 110: Should we be using [`os.Setenv`](https://golang.org/pkg/os/#Setenv) here?

cmd/kubectl.go
```diff
@@ -74,3 +117,75 @@ func runKubectl(ctx *cli.Context) error {
...
+}
+
+func extractKubeconfigTokenID(kubeconfig []byte) (string, error) {
+	type Token struct {
+		Users []struct {
+			User struct {
+				T string `yaml:"token"`
+			} `yaml:"user"`
+		} `json:"users,omitempty" yaml:"users,omitempty"`
+	}
```
aiyengar2, line 164: Should this be declared outside of the function?

cmd/server.go
```diff
@@ -64,15 +64,13 @@ be displayed and one can be selected.
 }
 
 // serverCurrent command to display the name of the current server in the local config
-func serverCurrent(ctx *cli.Context) error {
-	cf, err := loadConfig(ctx)
+func printServerCurrent(ctx *cli.Context) error {
```
aiyengar2, line 67:
```suggestion
func printCurrentServer(ctx *cli.Context) error {
```

aiyengar2 (changes requested):

cmd/kubectl.go
rmweir: changed to use HOME env var

cmd/server.go
rmweir: I don't think so. I was really just extracting some logic to the common.go, but I don't see the need to change printing format that is unrelated to the ticket since it's not objectively worse or anything.

cmd/kubectl.go
rmweir: Changed but is there any benefit of using that instead when the number of strings is already known apart from it being in that path package?

cmd/kubectl.go
rmweir: The format is only used once so it could be but is there a benefit?

cmd/kubectl.go
rmweir: I think this way may be better since the var is only needed for the single command and this contains it.

cmd/kubectl.go
aiyengar2: I think it's a better practice because it takes care of weird edge cases (e.g. if `home`, `kubeconfigpath`, or `filename` have trailing slashes or something similar).

cmd/kubectl.go
aiyengar2: Hmm not sure. I like to keep these as constants on the top even if it's used once since I don't like magic strings, but I'm fine with either decision!

cmd/kubectl.go
aiyengar2: Oh sorry you're right. I didn't realize this was just being set for this command.

cmd/kubectl.go
rmweir: I considered this as well, I decided to contain it within the function because it's strictly being used to unmarshal, and only to extract the token at that so idk if it would ever have a user outside of the function since it is incomplete.

rmweir:
@aiyengar2 @nickgerace respond to and or addressed feedback. Fixed a few additional things I noticed as well.

cmd/kubectl.go
aiyengar2: I see. Is there any way we could use the structs defined in [`client-go/tools/clientcmd`](https://github.com/kubernetes/client-go/blob/30548acd0a9e231d42d406433aef55b111b36240/tools/clientcmd/api/v1/types.go#L28-L52) instead?

We could use [LoadFromFile](https://godoc.org/k8s.io/client-go/tools/clientcmd#LoadFromFile), [IsEmptyConfig](https://godoc.org/k8s.io/client-go/tools/clientcmd#IsEmptyConfig), and [WriteToFile](https://godoc.org/k8s.io/client-go/tools/clientcmd#WriteToFile). This might allow us to eliminate / simplify `loadKubeConfig`, `writeKubeConfig`, and `extractKubeconfigTokenID` here without requiring an additional inline struct.

cmd/server.go
nickgerace: Ok sounds good

nickgerace (dismissed):

cmd/kubectl.go
rmweir: sure, I can do- will need to add client-go as a dependency though.

cmd/kubectl.go
```diff
@@ -74,3 +116,37 @@ func runKubectl(ctx *cli.Context) error {
...
+	return clientcmd.WriteToFile(*config, filePath)
+}
+
+func extractKubeconfigTokenID(kubeconfig api.Config) (string, error) {
+	if len(kubeconfig.AuthInfos) != 1 {
+		return "", fmt.Errorf("invalid kubeconfig, expected to contain exactly 1 user")
+	}
+	var parts []string
+	for _, val := range kubeconfig.AuthInfos {
```
dramich, line 133: No need for the loop with the check for 1 above

cmd/kubectl.go
```diff
@@ -37,34 +50,63 @@ func runKubectl(ctx *cli.Context) error {
...
 	if err != nil {
 		return err
 	}
-	defer os.Remove(tmpfile.Name())
 
-	_, err = tmpfile.Write([]byte(config.Config))
+	currentUser := t.UserID
+	currentCluster := currentRancherServer.FocusedCluster()
+	filename := fmt.Sprintf(kubeConfigNameFormat, currentUser, currentCluster)
+	kubeconfigFilePath := strings.Join([]string{kubeConfigPath, filename}, "/")
```
dramich, line 72: Use path.Join here

cmd/kubectl.go
```diff
@@ -2,11 +2,24 @@ package cmd
...
 	"fmt"
-	"io/ioutil"
 	"os"
 	"os/exec"
+	"strings"
 
 	"github.com/urfave/cli"
+	"k8s.io/client-go/tools/clientcmd"
+	"k8s.io/client-go/tools/clientcmd/api"
+
```
dramich, line 12: Update the sorting here, delete the blank and sort again

cmd/server.go
```diff
@@ -64,15 +64,13 @@ be displayed and one can be selected.
 }
 
 // serverCurrent command to display the name of the current server in the local config
-func serverCurrent(ctx *cli.Context) error {
-	cf, err := loadConfig(ctx)
+func printCurrentServer(ctx *cli.Context) error {
```
dramich, line 67: Comment should get updated as well

cmd/kubectl_token.go
```diff
@@ -51,7 +51,7 @@ type LoginInput struct {
 }
 
 const (
-	kubeConfigCache = "/.cache/token"
+	tokenCache      = "/.cache/token"
```
dramich, line 54: None of the changes in here are needed, this is getting changed in another pr

cmd/kubectl.go
rmweir: Spoke elsewhere, this is being used because the key is unknown.

cmd/kubectl.go
```diff
@@ -37,29 +47,80 @@ func runKubectl(ctx *cli.Context) error {
...
 	}
 
-	cluster, err := getClusterByID(c, c.UserConfig.FocusedCluster())
+	config, err := loadConfig(ctx)
 	if err != nil {
 		return err
 	}
 
-	config, err := c.ManagementClient.Cluster.ActionGenerateKubeconfig(cluster)
+	currentRancherServer := config.FocusedServer()
```
nickgerace, line 55: Should this function return `(server, err)` instead of just `server`? Might be a good change, but could be out of scope for this PR.

nickgerace (dismissed):
LGTM barring one nit

config/config.go
```diff
@@ -21,12 +22,13 @@ type Config struct {
...
-	URL       string `json:"url"`
-	Project   string `json:"project"`
-	CACerts   string `json:"cacert"`
+	AccessKey   string                 `json:"accessKey"`
+	SecretKey   string                 `json:"secretKey"`
+	TokenKey    string                 `json:"tokenKey"`
+	URL         string                 `json:"url"`
+	Project     string                 `json:"project"`
+	CACerts     string                 `json:"cacert"`
+	KubeConfigs map[string]*api.Config `json:"kubeConfigs"`
```
aiyengar2, line 31: nit: should this have the YAML fields as well or is the server config usually strictly JSON?

cmd/common.go
```diff
@@ -139,6 +143,40 @@ func listRoleTemplateBindings(ctx *cli.Context, b []RoleTemplateBinding) error {
...
+func setKubeConfigForUser(ctx *cli.Context, user string, kubeConfig *api.Config) error {
+	cf, err := loadConfig(ctx)
+	if err != nil {
+		return err
+	}
+
+	if cf.FocusedServer().KubeConfigs == nil {
+		cf.FocusedServer().KubeConfigs = make(map[string]*api.Config)
+	}
+	cf.FocusedServer().KubeConfigs[fmt.Sprintf(kubeConfigKeyFormat, user, cf.FocusedServer().FocusedCluster())] = kubeConfig
```
aiyengar2, line 176: nit: should `cf.FocusedServer()` be just one call that gets stored in a variable?

cmd/common.go
```diff
@@ -139,6 +143,40 @@ func listRoleTemplateBindings(ctx *cli.Context, b []RoleTemplateBinding) error {
...
+	return cf.CurrentServer, cf.Servers[serverName].URL, nil
+}
+
+func getCurrentKubeConfigForUser(ctx *cli.Context, user string) (*api.Config, error) {
+	cf, err := loadConfig(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	kubeConfig, _ := cf.FocusedServer().KubeConfigs[fmt.Sprintf(kubeConfigKeyFormat, user, cf.FocusedServer().FocusedCluster())]
```
aiyengar2, line 163: nit: should `cf.FocusedServer()` be just one call that gets stored in a variable?

cmd/kubectl.go
```diff
@@ -37,29 +47,80 @@ func runKubectl(ctx *cli.Context) error {
...
+		if err != nil {
+			return err
+		}
+		isTokenValid, err = validateToken(tokenID, c.ManagementClient.Token)
+		if err != nil {
+			return err
+		}
+	}
+
+	var kubeconfigBytes []byte
```
aiyengar2, line 84: nit: `kubeconfigBytes` is inconsistent. Should be `kubeConfigBytes`

aiyengar2 (dismissed):
Small comments, overall look good to me.

cmd/kubectl.go
rmweir: I don't think so, the function just accesses a map.

config/config.go
rmweir: This is how it's been, I only added a field. I think most YAML libraries use the json tags now.

cmd/common.go
rmweir: It's just access a map so it's not an expensive call, but it would probably be more readable, will change.

rmweir: Using a variable doesn't work for L174 since I want to assign a new values to one of it's fields.

aiyengar2 (dismissed):
LGTM

nickgerace (dismissed):

cmd/kubectl.go
```diff
@@ -37,29 +43,80 @@ func runKubectl(ctx *cli.Context) error {
...
+			return err
+		}
+	}
+
+	if len(kubeConfigBytes) == 0 {
+		kubeConfigBytes, err = yaml.Marshal(kubeConfig)
+		if err != nil {
+			return err
+		}
+	}
```
dramich, line 108: kubeConfigBytes is not used after this so probably not needed

cmd/common.go
```diff
@@ -139,6 +143,42 @@ func listRoleTemplateBindings(ctx *cli.Context, b []RoleTemplateBinding) error {
...
+	cf, err := loadConfig(ctx)
+	if err != nil {
+		return "", "", err
+	}
+
+	serverName := cf.CurrentServer
+	return cf.CurrentServer, cf.Servers[serverName].URL, nil
+}
+
+func getCurrentKubeConfigForUser(ctx *cli.Context, user string) (*api.Config, error) {
```
dramich, line 157:
```suggestion
func getKubeConfigForUser(ctx *cli.Context, user string) (*api.Config, error) {
```

cmd/common.go
```diff
@@ -139,6 +143,42 @@ func listRoleTemplateBindings(ctx *cli.Context, b []RoleTemplateBinding) error {
 	return writer.Err()
 }
 
+// getCurrentServer returns the name of the name and url of the current server in the local config
```
dramich, line 146:
```suggestion
// getCurrentServer returns the name and url of the current server in the local config
```

cmd/server.go
```diff
@@ -63,16 +63,14 @@ be displayed and one can be selected.
...
+	serverName, url, err := getCurrentServer(ctx)
 	if err != nil {
 		return err
 	}
 
-	serverName := cf.CurrentServer
-	URL := cf.Servers[serverName].URL
-	fmt.Printf("Name: %s URL: %s\n", serverName, URL)
+	fmt.Printf("Name: %s URL: %s\n", serverName, url)
```
dramich, line 73: What is the purpose of this change? It doesn't seem related to the rest of the PR

cmd/common.go
```diff
@@ -139,6 +143,42 @@ func listRoleTemplateBindings(ctx *cli.Context, b []RoleTemplateBinding) error {
...
+func getCurrentServer(ctx *cli.Context) (string, string, error) {
```
dramich, line 147: I don't think this is needed, it's only used in one place not related to the pr

dramich (changes requested):

cmd/server.go
rmweir: I set this up prior to moving to the config approach. I wanted to put it in common to be accessed elsewhere. No longer needed, will remove.

nickgerace (dismissed):

cmd/kubectl.go
```diff
@@ -37,29 +42,73 @@ func runKubectl(ctx *cli.Context) error {
...
+		if err != nil {
+			return err
+		}
+		isTokenValid, err = validateToken(tokenID, c.ManagementClient.Token)
+		if err != nil {
+			return err
+		}
+	}
+
+	var kubeConfigBytes []byte
```
dramich, line 79: This is only used inside the if block so it probably doesn't need to be created here

dramich (approved):

PR #294, opened 2020-12-04
minor fixes for kubeconfig exec credential

kinarashah:
- set correct kind for ExecCredential
- update help command to reflect Rancher config dir
- check for nil timestamp


dramich (approved):

dramich:
https://github.com/rancher/rancher/issues/30092

PR #296, opened 2020-12-18
Automatically open browser when getting token

tjcain:
Small QOL improvement which automatically opens the browser during `rancher token` if an oidc provider is selected.


mateuszdrab:
Hey @tjcain

I was wondering if this PR was supposed to make the CLI work with Azure AD since it's using OIDC...

Currently, latest version of the CLI throws an error when trying to use the AAD provider:
`FATA[0004] error logging in: code: [NotFound] message:[failed to find schema azureads]`

I don't think this PR brings a fix to that - what OIDC provider did you have in mind?

Thanks

PR #305, opened 2021-09-27
Write only base-names of files in the checksum file on releases

genexpr:
Issue [#33640](https://github.com/rancher/rancher/issues/33640)


StrongMonkey (approved):

StrongMonkey:
@genexpr you can tag an new rc in CLI and check the [sha256sum.txt](https://github.com/rancher/cli/releases/download/v2.4.13/sha256sum.txt) file. It should has the basename of the file.

genexpr:
@StrongMonkey, that means after merging I should create a new pre-release with a tag of `v2.4.14-rc1`, right?

StrongMonkey:
@genexpr That's right

superseb (changes requested):
Part of this should be making sure we do this for all of our repositories or not I think, currently they all not use this but just changing one causing discrepancy with the other repos and if we decide to change this, we should change it for all repositories that publish checksum so we know what to expect. Besides users having to change their checksum command, I don't think anything else is broken by this change?

genexpr:
@superseb, good point! One question: it does not matter what kind of assets we release, right? For example, in `rancher/rancher`, there are no binaries, just scripts and text files. But users still might want to run a checksum to verify the integrity of those files?

Also, should we first merge this one, then make a pre-release RC - just to test it? If drone processes the paths as we want, then we can add the same change to the other repos? By the way, what exact other repos besides the main `rancher` one? Every one that has releases conducted by drone?

genexpr:
We use this flag in a few other `rancher` repositories, such as `rke2`, `terraform-controller`, and about 30 others. Which is good, because there is no inconsistency, file path flattening for a checksum file is used where it's needed.

StrongMonkey (approved):

PR #306, opened 2021-10-11
fix: default config file path on Windows

philomory:
This fixes rancher/rancher#31967 (which was closed by the reporter, but is definitely still an issue). Basically, use `filepath.Join()` rather than assuming `/` is the path separator.

### DIFF
--- a/cmd/kubectl_token.go
+++ b/cmd/kubectl_token.go
@@ -17,6 +17,7 @@ import (
 	url2 "net/url"
 	"os"
 	"os/signal"
+	"path/filepath"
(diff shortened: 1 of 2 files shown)


chrismeller:
+1, as this fixed my problem.

rahulmr:
Please fix this. Sometimes it feels like there is issue from our side. Is there any documentation in the CLI github repo for exporting the variable on windows?

HarrisonWAffel:
Closed in favor of https://github.com/rancher/cli/pull/326

PR #310, opened 2021-11-24
[DNM until 2.6.3 is out] Transform all keys to JSON format for RKE config values

genexpr:
Currently, the CLI does not honor most values in a passed in `rke-config` file. The CLI silently ignored it, and new clusters used default values.

This fixes it, so the CLI respects and applies the settings in `rancher_kubernetes_engine_config`.
The same is now true for top-level keys above it, like `docker_root_dir`.
Unfortunately, this cannot be a perfect solution. But it's better than what we have now.

The problem is that most values in `RancherKubernetesEngineConfig` are defined with struct tags for both JSON and YAML in camelCase.
https://github.com/rancher/rancher/blob/release/v2.6/pkg/client/generated/management/v3/zz_generated_rancher_kubernetes_engine_config.go

Changing the tags will be a breaking change. For proper deserialization, we must convert all keys to camelCase.
Note that we ignore kebab-case keys. Users themselves should ensure any relevant keys
(especially top-level keys in `services`, like `kube-api` or `kube-controller`) are camelCase or snake-case in cluster config.
Otherwise, they will be ignored on deserialization, and defaults settings will be used. And some settings in `kube-api`, for example, are expected to be in kebab-case.

Our RKE docs, unfortunately, use kebab-case for some examples.

This should also fix the `import` flag. It would not do anything previously and create a default cluster. Now it just defers to the import command handler.

I would love to write an end-to-end test for this workflow: supply an RKE config file, compare resultant cluster configuration.
But we currently have no testing framework in place for such things in the CLI repository, since every such workflow makes a real API request to a live cluster.

### DIFF
--- a/cmd/cluster.go
+++ b/cmd/cluster.go
@@ -250,6 +250,9 @@ func clusterCreate(ctx *cli.Context) error {
 	if ctx.NArg() == 0 {
 		return cli.ShowSubcommandHelp(ctx)
 	}
+	if ctx.Bool("import") {
+		return clusterImport(ctx)
+	}
 	c, err := GetClient(ctx)
 	if err != nil {
 		return err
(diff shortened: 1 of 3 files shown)


genexpr:
This one is ready for review, now that v2.6.3 is out, @superseb, @rawmind0. Thanks!

paynejacob (approved):

kinarashah (approved):
Thanks for adding tests, make sure QA looks out for the vsphere virtual_center field mentioned in the original issue.

PR #316, opened 2022-05-24
Display message when no servers are available

HarrisonWAffel:
Issue: https://github.com/rancher/rancher/issues/37545

# Problem
When running `rancher-cli servers ls`, `rancher-cli servers delete` or `rancher-cli servers switch` when there are no servers configured, the cli will still prompt the user to select a server 

# Solution
Check if any servers exist within the configuration before prompting the user. If no servers are stored in the configuration, print a message to the user and do not prompt for further input. 

# Testing
delete the  `cli2.json` file to clear out any server configurations 
Run `rancher-cli servers ls` and observe that it does not prompt you for input 
Run `rancher-cli servers delete` and observe that it does not prompt you for input 
Run `rancher-cli servers switch` and observe that it does not prompt you for input

### DIFF
--- a/cmd/server.go
+++ b/cmd/server.go
@@ -83,6 +83,10 @@ func serverDelete(ctx *cli.Context) error {
 		return err
 	}
 
+	if err := validateServersConfig(cf); err != nil {
+		return err
+	}
+
 	var serverName string
 	if ctx.NArg() == 1 {
 		serverName = ctx.Args().First()
@@ -114,6 +118,10 @@ func serverLs(ctx *cli.Context) error {
 		return err
 	}
 
+	if err := validateServersConfig(cf); err != nil {
+		return err
+	}
+
 	writer := NewTableWriter([][]string{
 		{"CURRENT", "Current"},
 		{"NAME", "Name"},
@@ -141,6 +149,10 @@ func serverLs(ctx *cli.Context) error {
 func serverSwitch(ctx *cli.Context) error {
 	cf, err := loadConfig(ctx)
 	if err != nil {
+		return err
+	}
+
+	if err := validateServersConfig(cf); err != nil {
 		return err
 	}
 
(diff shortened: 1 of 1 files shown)


snasovich (approved):
Changes look good to me overall.
Contemplated asking to extract a separate function for validating there are configured servers, but the benefits are minimal (if any) so an argument could be made for either approach being OK.

Meanwhile, could you please look what's wrong with builds, it's complaining about `golangci-lint` not being available - certainly not related to this PR but needs to be resolved first.

cmd/server.go
```diff
@@ -83,6 +83,11 @@ func serverDelete(ctx *cli.Context) error {
 		return err
 	}
 
+	if len(cf.Servers) == 0 {
+		fmt.Println("No servers are currently configured.")
+		return nil
+	}
```
jakefhyde, line 89:
```suggestion
	if len(cf.Servers) == 0 {
		return errors.New("no servers are currently configured")
	}
```

jakefhyde (changes requested):

cmd/server.go
snasovich: Good catch, @jakefhyde.

If we're doing this, I'd say it puts more weight into separating it into a func and doing something like below in all 3 places:
```
	if err := validateServersConfig(cf); err != nil {
		return err
	}
```

HarrisonWAffel: If we return an error the resulting log message is `FATA[0000] No servers are currently configured. `. I'm not sure if we want to include `FATA[0000]` when this occurs, as it's not necessarily a fatal _error_ and does not add any additional information to the person running the command. Adding the `FATA[0000]` is done by `github.com/urfave/cli`, so I cannot suppress that section of the log message if we do return an error. 

Do we still want to return an error in this case? Not have any servers configured isn't necessarily an error, although I do understand the sentiment that because this command did not result in the described action it could be considered an error. I'm fine with either decision, but I wanted to note the differences

snasovich: We already return errors from this func, though these errors have very low probability of actually happening.
Regardless, I agree with Jake that in case of no servers configured, we have error condition and CLI should not exit with 0.

HarrisonWAffel:
@snasovich these build failures are known and a fix is pending, https://github.com/rancher/tasks/issues/203. Downloading the linter is failing, but I've made sure that this error message will pass the linter once it has been fixed.

PR #318, opened 2022-06-07
Fix index out of bounds when installing hidden chart

PennyScissors:
Issue: https://github.com/rancher/rancher/issues/36332

Problem:
Charts in the old Rancher catalog are deprecated and hidden from the user. When you attempt to install any of them the Rancher cli tries to find the latest version of the chart and panics with an index out of bounds.

Solution:
This PR adds a conditional to recover from the panic gracefully, and provide an error with a description of the problem. Note this is an edge case and will most likely only be seen by users with the legacy flag enabled trying to use deprecated charts.

### DIFF
--- a/cmd/app.go
+++ b/cmd/app.go
@@ -1034,6 +1034,9 @@ func getFilteredTemplate(ctx *cli.Context, c *cliclient.MasterClient, templateID
 
 // getTemplateLatestVersion returns the newest version of the template
 func getTemplateLatestVersion(template *managementClient.Template) (string, error) {
+	if len(template.VersionLinks) == 0 {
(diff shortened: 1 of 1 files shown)


superseb (commented):
Based on the issue, we are just fixing the panic and not providing an error that explains why this doesn't work? We already hide the v1 apps right, so we can also state in the error that it's no longer supported?

I guess what I am trying to say is, we are fixing the panic but the user still doesn't know what's going on from the error given.

PennyScissors:
@superseb I updated the error message to provide more information about what the problem is exactly.

superseb (approved):

PR #322, opened 2022-08-09
Update error messages and usage for add-node command

HarrisonWAffel:
Issue: https://github.com/rancher/rancher/issues/38515

This PR updates the error messages and usage of the `clusters add-node` command so that there is less ambiguity on its proper usage.


jakefhyde (dismissed):

annablender (dismissed):

jakefhyde (approved):

annablender (approved):
Logic looks good

PR #325, opened 2022-09-27
Assume cluster-id as argument instead of either cluster name or id

steffantucker:
Updates cluster commands to require an explicit flag if using a cluster name instead of a cluster id. This could be a breaking change, as previously you could pass either the id or name.

### DIFF
--- a/cmd/cluster.go
+++ b/cmd/cluster.go
@@ -8,6 +8,7 @@ import (
 	"strings"
 
 	"github.com/rancher/cli/cliclient"
+	"github.com/rancher/norman/types"
 	managementClient "github.com/rancher/rancher/pkg/client/generated/management/v3"
 	"github.com/sirupsen/logrus"
 	"github.com/urfave/cli"
@@ -98,18 +99,26 @@ func ClusterCommand() cli.Command {
 				Name:        "import",
 				Usage:       "Import an existing Kubernetes cluster into a Rancher cluster",
 				Description: importDescription,
-				ArgsUsage:   "[CLUSTERID CLUSTERNAME]",
+				ArgsUsage:   "CLUSTERID|--cluster-name CLUSTERNAME",
 				Action:      clusterImport,
 				Flags: []cli.Flag{
+					cli.StringFlag{
+						Name:  "cluster-name, name, n",
+						Usage: "Specify cluster by `CLUSTERNAME` instead of CLUSTERID",
+					},
 					quietFlag,
 				},
 			},
 			{
 				Name:      "add-node",
 				Usage:     "Outputs the docker command needed to add a node to an existing Rancher custom cluster",
-				ArgsUsage: "[CLUSTERID CLUSTERNAME]",
+				ArgsUsage: "CLUSTERID|--cluster-name CLUSTERNAME",
 				Action:    clusterAddNode,
 				Flags: []cli.Flag{
+					cli.StringFlag{
+						Name:  "cluster-name, name, n",
+						Usage: "Specify cluster by `CLUSTERNAME` instead of CLUSTERID",
+					},
 					cli.StringSliceFlag{
 						Name:  "label",
 						Usage: "Label to apply to a node in the format [name]=[value]",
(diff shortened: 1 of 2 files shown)


cmd/common.go
```diff
@@ -430,6 +385,102 @@ func Lookup(c *cliclient.MasterClient, name string, types ...string) (*ntypes.Re
...
+	}
+
+	if len(collection.Data) > 1 {
+		ids := []string{}
+		for _, data := range collection.Data {
+			ids = append(ids, data.ID)
+		}
+		return nil, fmt.Errorf("Multiple resources of type %s found for name %s: %v", schemaType, name, ids)
+	}
+	if len(collection.Data) == 0 {
```
maxsokolovsky, line 454: Should we return a "Not Found" error here as well?

steffantucker: it seems like i had a reason not to, but can't see it now.. so I'll ad a not found return

cmd/cluster.go
```diff
@@ -143,15 +152,27 @@ func ClusterCommand() cli.Command {
...
 				Name:      "kubeconfig",
 				Aliases:   []string{"kf"},
 				Usage:     "Return the kube config used to access the cluster",
-				ArgsUsage: "[CLUSTERID CLUSTERNAME]",
+				ArgsUsage: "[CLUSTERID]",
 				Action:    clusterKubeConfig,
+				Flags: []cli.Flag{
+					cli.StringFlag{
+						Name:  "cluster-name, name, n",
+						Usage: "Get kube config for `CLUSTERNAME`",
```
maxsokolovsky, line 173: Nit: `kubeconfig`. Could be changed in a couple of other places in usages.

cmd/common.go
```diff
@@ -348,78 +349,32 @@ func Lookup(c *cliclient.MasterClient, name string, types ...string) (*ntypes.Re
...
-			},
+		if resource != nil {
+			return resource, nil
 		}
 
-		if err := schemaClient.List(schemaType, listOpts, &collection); !clientbase.IsNotFound(err) && err != nil {
-			logrus.Debugf("Error schemaClient.List: %v", err)
+		// Resource was not found assuming the ID, check if it's the name of a resource
+		resource, err = LookupByName(c, name, schemaType)
+		if err != nil && !strings.Contains(err.Error(), "Not found") {
```
KevinJoiner, line 363: We should always try and avoid doing string checks on `err.Error()` and instead do a type check. 
To accomplish this, we will need to:
-  define `errNotFound` as a package variable.
- Return `errNotFound` around line 436.
- Check if `err == errNotFound`

KevinJoiner (changes requested):
Did you mean to leave the delete command the same on line 140?

cmd/cluster.go
```diff
@@ -98,18 +99,26 @@ func ClusterCommand() cli.Command {
 				Name:        "import",
 				Usage:       "Import an existing Kubernetes cluster into a Rancher cluster",
 				Description: importDescription,
-				ArgsUsage:   "[CLUSTERID CLUSTERNAME]",
+				ArgsUsage:   "[CLUSTERID]",
```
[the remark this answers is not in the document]
steffantucker: I don't think you're wrong, but this is how all the rest of the ArgsUsage are formatted for all other commands, which is why I had left the brackets

cmd/cluster.go
steffantucker: changed

cmd/cluster.go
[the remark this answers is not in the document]
steffantucker: so i did end up changing this for the commands i updated, to reflect that it's either the ID or the name with the flag

cmd/cluster.go
```diff
@@ -98,18 +99,26 @@ func ClusterCommand() cli.Command {
...
 			{
 				Name:      "add-node",
 				Usage:     "Outputs the docker command needed to add a node to an existing Rancher custom cluster",
-				ArgsUsage: "[CLUSTERID CLUSTERNAME]",
+				ArgsUsage: "[CLUSTERID]",
 				Action:    clusterAddNode,
 				Flags: []cli.Flag{
+					cli.StringFlag{
+						Name:  "cluster-name, name, n",
+						Usage: "Add node to `CLUSTERNAME` instead of CLUSTERID",
```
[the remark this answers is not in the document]
steffantucker: this is fair, updated all usages

maxsokolovsky (approved):
LGTM! Does this go into a CLI release that coincides with v2.7.0? And would there be a back port for the v2.6 branch of the CLI?

cmd/common.go
```diff
@@ -344,92 +346,157 @@ func GetResourceType(c *cliclient.MasterClient, resource string) (string, error)
...
 		}
 
-		byName = &collection.Data[0]
+		byName = resource
 
 	}
 
 	if byName == nil {
-		return nil, fmt.Errorf("Not found: %s", name)
+		return nil, fmt.Errorf("%w: %s", errNotFound, name)
```
KevinJoiner, line 388:
```suggestion
		return nil, fmt.Errorf("%s: %w", name, newRole)
```
Wrapping convention is to have the wrapped error at the end prefixed by a colon.

cmd/common.go
```diff
@@ -344,92 +346,157 @@ func GetResourceType(c *cliclient.MasterClient, resource string) (string, error)
...
+
+	if err := schemaClient.List(schemaType, listOpts, &collection); err != nil {
+		logrus.Debugf("Error schemaClient.List: %v", err)
+		if clientbase.IsNotFound(err) {
+			return nil, fmt.Errorf("%w: %s", errNotFound, name)
+		}
+		// prevents 403 from falsely returning as an unhandled error
+		if apiError, ok := err.(*clientbase.APIError); ok {
+			if apiError.StatusCode == http.StatusForbidden {
+				return nil, fmt.Errorf("%w: %s", errNotFound, name)
```
KevinJoiner, line 452:
```suggestion
				return nil, fmt.Errorf("%s: %w", name, errNotFound)
```

cmd/common.go
```diff
@@ -344,92 +346,157 @@ func GetResourceType(c *cliclient.MasterClient, resource string) (string, error)
...
+
+	if len(collection.Data) > 1 {
+		ids := []string{}
+		for _, data := range collection.Data {
+			ids = append(ids, data.ID)
+		}
+		return nil, fmt.Errorf("multiple resources of type %s found for name %s: %v", schemaType, name, ids)
+	}
+	if len(collection.Data) == 0 {
+		return nil, fmt.Errorf("%w: %s", errNotFound, name)
```
KevinJoiner, line 466: same as above

cmd/common.go
```diff
@@ -344,92 +346,157 @@ func GetResourceType(c *cliclient.MasterClient, resource string) (string, error)
...
+	return lookupByIDWithClient(schemaClient, id, schemaType)
+}
+
+func lookupByIDWithClient(schemaClient clientbase.APIBaseClientInterface, id, schemaType string) (*ntypes.Resource, error) {
+	var resource ntypes.Resource
+
+	if err := schemaClient.ByID(schemaType, id, &resource); err != nil {
+		logrus.Debugf("Error schemaClient.ByID: %v", err)
+		if clientbase.IsNotFound(err) {
+			return nil, fmt.Errorf("%w: %s", errNotFound, id)
```
KevinJoiner, line 487: same as above

cmd/common.go
```diff
@@ -344,92 +346,157 @@ func GetResourceType(c *cliclient.MasterClient, resource string) (string, error)
...
+		}
+		// prevents 403 from falsely returning as an unhandled error
+		if apiError, ok := err.(*clientbase.APIError); ok {
+			if apiError.StatusCode == http.StatusForbidden {
+				return nil, fmt.Errorf("%w: %s", errNotFound, id)
```
KevinJoiner, line 492: same as above

KevinJoiner (changes requested):
Should we also look at test?

KevinJoiner (approved):

maxsokolovsky (approved):

PR #326, opened 2022-11-14
fix: make the cli runnable on windows

bakito:
As as already mentioned in https://github.com/rancher/cli/pull/306, the cli is not running properly in windows, as the parent dir of the config file is not evaluated correctly. 

This PR fixes this and uses also `os.UserHomeDir()` to evaluate the home dir of the user instead of just expanding the HOME variable.


HarrisonWAffel (dismissed):
LGTM

Thanks for the contribution

HarrisonWAffel (approved):

PR #328, opened 2022-12-20
Config store helper

gregod-com:
This MR allows to define a local executable via the flag `--config-helper` or EnvVar `RANCHER_CONFIG_HELPER` to store the rancher config.

The methods to `load` and `write` the config are intercepted and by default keep the existing method of storing the config as file on the filesystem. (~/.rancher/cli2.json)
If the flag `--config-helper` is set to a other values than `built-in`, the method invokes a call to the defined executable with the argument `get` to load the config and `store` to persist the config.

An example config-helper could look like this:

rancher-helper
```
#!/bin/zsh

case "$1" in
    get)   gopass show -o -y rancher-config ;;
    store) echo $2 | gopass insert -f rancher-config ;;
esac
```

The motivation for this MR is to allow a secure storage of the credentials found in the rancher config.

### DIFF
--- a/cmd/common.go
+++ b/cmd/common.go
@@ -263,7 +263,7 @@ func verifyCert(caCert []byte) (string, error) {
 
 func loadConfig(ctx *cli.Context) (config.Config, error) {
 	switch ctx.GlobalString("config-helper") {
-	case "build-in":
+	case "built-in":
 		return loadConfigNative(ctx)
 	default:
 		// allow loading of rancher config by triggering an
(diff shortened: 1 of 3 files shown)


gregod-com:
> hi @gregod-com sorry for the delay - is this still something you would like to pursue?

Hey @samjustus,

yeah sure. As far as I remember the change was quite small and straight forward. I did/do not have the time to keep a fork of rancher cli open and in sync, bit it should be easy to recreate the PR with my changes. I can have a look in the coming days, if somebody is willing to review.

It would definitely be nice to not have credentials just in clear text on the hard disk anymore 😀

samjustus:
@gregod-com thanks - we find the new PR promising! Currently our QA team is swamped so we will have to see if we can fit this in right now but indeed we will review and go from there

PR #332, opened 2023-03-02
Fix import flag for cluster create command

daniel-swe:
Fixes the `cluster create --import` command trying to import a non-existent cluster instead of creating a generic cluster for import.
First time around this project, let me know if I'm missing something!

Tracked at [rancher#37574](https://github.com/rancher/rancher/issues/37574) and https://github.com/rancher/rancher/issues/41144

### DIFF
--- a/cmd/cluster.go
+++ b/cmd/cluster.go
@@ -277,9 +277,6 @@ func clusterCreate(ctx *cli.Context) error {
 	if ctx.NArg() == 0 {
 		return cli.ShowSubcommandHelp(ctx)
 	}
-	if ctx.Bool("import") {
-		return clusterImport(ctx)
-	}
 	c, err := GetClient(ctx)
 	if err != nil {
 		return err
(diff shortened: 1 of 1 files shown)


bfbachmann (approved):
Thank you! LGTM

cmd/cluster.go
```diff
@@ -277,9 +277,6 @@ func clusterCreate(ctx *cli.Context) error {
 	if ctx.NArg() == 0 {
 		return cli.ShowSubcommandHelp(ctx)
 	}
-	if ctx.Bool("import") {
-		return clusterImport(ctx)
-	}
```
KevinJoiner, line 282: Why are we removing this? It looks like `clusterImport(...)` is unused now

KevinJoiner (changes requested):

cmd/cluster.go
daniel-swe: I think `clusterImport(...)` is already used for the `cluster import` subcommand [here](https://github.com/daniel-swe/rancher-cli/blob/f5d2d57288a089081da9e8f0c690f911b27acfd5/cmd/cluster.go#L103C5-L103C5).

bfbachmann: I had the same thought initially, but this function is still used in the `rancher cluster import` command. That function is actually only supposed to be called when the user is doing `rancher cluster import`, as it actually calls Rancher to generate a token and prints out the `kubectl` command to run on the downstream cluster to import it to Rancher. The fact that it's called here is actually the main cause of the bug that is being fixed in this PR.

This function `clusterCreate` is called when a user does `rancher cluster create [--import]`, and it calls Rancher to create the cluster objects, so it shouldn't actually be calling `clusterImport` when the `--import` flag is used. Instead, if the `--import` flag is passed, it should simply be calling Rancher to create "empty" cluster objects to use as a target when the user later runs `rancher cluster import`. That's what this PR is doing. 

It's very confusing because of the way the functions and commands are named.

bfbachmann: @KevinJoiner lmk if that makes sense. It's quite possible that I'm wrong.

bfbachmann: I think the original bug probably happened because someone confused the functionality in `rancher cluster create --import` with the functionality in `rancher cluster import`.

KevinJoiner: Thanks for the explanation I looked at this [PR](https://github.com/rancher/cli/pull/310/files), and I think you are right.

KevinJoiner (approved):

PR #334, opened 2023-03-28
Allow insecure server connections

wjiec:
Now, add a `--insecure` option to skip the SSL certificate validation.

When the SSL certificate used by the rancher server is invalid, it will prompt `FATA[0000] Get "https://rancher.example.com/v3": x509: “dynamic” certificate is not trusted`.


samjustus:
hi @wjiec ... sorry for the delay - is this still something you would like to pursue?

PR #341, opened 2023-09-19
Create config with 0600 permissions by default

tomleb:
Issue: rancher/rancher#42838

# Problem

The default permissions was `0644`  which is not the most secure default.

# Solution

Change the default permission to `0600`. Keep same permissions for config files that already exists. Warn the user if config file is either group-readable or world-readable.

# Summary of changes

First commit moves the bulk of the loading code to `config/` module, it seems more appropriate. I also added some tests to make sure I didn't totally break anything with the move.

Second commit does the following:
- Changed the default file creation to `0600` permissions. As discussed, already existing files won't have their permissions changed, this only affects new files.
- A warning is emitted if the file is either group-readable or world-readable. The message was heavily inspired by the kubectl one. I've made it so that the warning is only written once per CLI invocation, since some commands read the config multiple times.

The result is:

```
$ chmod o+r ~/.rancher/cli2.json
$ ./bin/rancher login --token $token https://example.com
WARN[0000] Rancher configuration file /home/tlebreux/.rancher/cli2.json is world-readable. This is insecure. 
INFO[0000] Saving config to /home/tlebreux/.rancher/cli2.json

$ chmod g+r ~/.rancher/cli2.json 
$ ./bin/rancher login --token $token https://example.com
WARN[0000] Rancher configuration file /home/tlebreux/.rancher/cli2.json is group-readable. This is insecure. 
WARN[0000] Rancher configuration file /home/tlebreux/.rancher/cli2.json is world-readable. This is insecure. 
INFO[0000] Saving config to /home/tlebreux/.rancher/cli2.json  

$ chmod 0600 ~/.rancher/cli2.json
$ ./bin/rancher login --token $token https://example.com
INFO[0000] Saving config to /home/tlebreux/.rancher/cli2.json 
```

# Checklist

I have tested done some manual tests:
- Made sure that existing file don't have their permission changed
- All permuttations of g+r, o+r emit the appropriate warnings
- Some commands still work (eg: login, nodes)
- Verified that the warning is outputted to stderr, not stdout (makes sure it doesn't break scripts that rely on outputs)

### DIFF
--- a/cmd/common.go
+++ b/cmd/common.go
@@ -5,7 +5,6 @@ import (
 	"bytes"
 	"context"
 	"crypto/x509"
-	"encoding/json"
 	"encoding/pem"
 	"fmt"
 	"io"
@@ -259,27 +258,15 @@ func verifyCert(caCert []byte) (string, error) {
 	return string(caCert), nil
 }
 
-func loadConfig(ctx *cli.Context) (config.Config, error) {
+func GetConfigPath(ctx *cli.Context) string {
 	// path will always be set by the global flag default
 	path := ctx.GlobalString("config")
-	path = filepath.Join(path, cfgFile)
+	return filepath.Join(path, cfgFile)
+}
 
-	cf := config.Config{
-		Path:    path,
-		Servers: make(map[string]*config.ServerConfig),
-	}
-
-	content, err := ioutil.ReadFile(path)
-	if os.IsNotExist(err) {
-		return cf, nil
-	} else if err != nil {
-		return cf, err
-	}
-
-	err = json.Unmarshal(content, &cf)
-	cf.Path = path
-
-	return cf, err
+func loadConfig(ctx *cli.Context) (config.Config, error) {
+	path := GetConfigPath(ctx)
+	return config.LoadFromPath(path)
 }
 
 func lookupConfig(ctx *cli.Context) (*config.ServerConfig, error) {
--- a/config/config.go
+++ b/config/config.go
@@ -59,6 +59,30 @@ func LoadFromPath(path string) (Config, error) {
 	return cf, nil
 }
 
+// GetFilePermissionWarnings returns the following warnings based on the file permission:
+// - one warning if the file is group-readable
+// - one warning if the file is world-readable
+// We want this because configuration may have sensitive information (eg: creds).
+// A nil error is returned if the file doesn't exist.
+func GetFilePermissionWarnings(path string) ([]string, error) {
+	info, err := os.Stat(path)
+	if err != nil {
+		if os.IsNotExist(err) {
+			return []string{}, nil
+		}
+		return []string{}, fmt.Errorf("get file info: %w", err)
+	}
+
+	var warnings []string
+	if info.Mode()&0040 > 0 {
+		warnings = append(warnings, fmt.Sprintf("Rancher configuration file %s is group-readable. This is insecure.", path))
+	}
+	if info.Mode()&0004 > 0 {
+		warnings = append(warnings, fmt.Sprintf("Rancher configuration file %s is world-readable. This is insecure.", path))
+	}
+	return warnings, nil
+}
+
 func (c Config) Write() error {
 	err := os.MkdirAll(path.Dir(c.Path), 0700)
 	if err != nil {
@@ -68,7 +92,7 @@ func (c Config) Write() error {
 	logrus.Infof("Saving config to %s", c.Path)
 	p := c.Path
 	c.Path = ""
-	output, err := os.Create(p)
+	output, err := os.OpenFile(p, os.O_RDWR|os.O_CREATE|os.O_TRUNC, 0600)
 	if err != nil {
 		return err
 	}
--- a/config/config.go
+++ b/config/config.go
@@ -2,6 +2,7 @@ package config
 
 import (
 	"encoding/json"
+	"fmt"
 	"net/url"
 	"os"
 	"path"
@@ -30,6 +31,32 @@ type ServerConfig struct {
 	CACerts         string                     `json:"cacert"`
 	KubeCredentials map[string]*ExecCredential `json:"kubeCredentials"`
 	KubeConfigs     map[string]*api.Config     `json:"kubeConfigs"`
+}
+
+// LoadFromPath attempts to load a config from the given file path. If the file
+// doesn't exist, an empty config is returned.
+func LoadFromPath(path string) (Config, error) {
+	cf := Config{
+		Path:    path,
+		Servers: make(map[string]*ServerConfig),
+	}
+
+	content, err := os.ReadFile(path)
+	if err != nil {
+		// it's okay if the file is empty, we still return a valid config
+		if os.IsNotExist(err) {
+			return cf, nil
+		}
+
+		return cf, err
+	}
+
+	if err := json.Unmarshal(content, &cf); err != nil {
+		return cf, fmt.Errorf("unmarshaling %s: %w", path, err)
+	}
+	cf.Path = path
+
+	return cf, nil
 }
 
 func (c Config) Write() error {
--- a/config/config_test.go
+++ b/config/config_test.go
@@ -28,6 +28,109 @@ const (
 	invalidFile = `invalid config file`
 )
 
+func Test_GetFilePermissionWarnings(t *testing.T) {
+	t.Parallel()
+
+	tests := []struct {
+		name             string
+		mode             os.FileMode
+		expectedWarnings int
+	}{
+		{
+			name:             "neither group-readable nor world-readable",
+			mode:             os.FileMode(0600),
+			expectedWarnings: 0,
+		},
+		{
+			name:             "group-readable and world-readable",
+			mode:             os.FileMode(0644),
+			expectedWarnings: 2,
+		},
+		{
+			name:             "group-readable",
+			mode:             os.FileMode(0640),
+			expectedWarnings: 1,
+		},
+		{
+			name:             "world-readable",
+			mode:             os.FileMode(0604),
+			expectedWarnings: 1,
+		},
+	}
+	for _, tt := range tests {
+		tt := tt
+		t.Run(tt.name, func(t *testing.T) {
+			t.Parallel()
+			assert := assert.New(t)
+
+			dir, err := os.MkdirTemp("", "rancher-cli-test-*")
+			assert.NoError(err)
+			defer os.RemoveAll(dir)
+
+			path := filepath.Join(dir, "cli2.json")
+			err = os.WriteFile(path, []byte(validFile), tt.mode)
+			assert.NoError(err)
+
+			warnings, err := GetFilePermissionWarnings(path)
+			assert.NoError(err)
+			assert.Len(warnings, tt.expectedWarnings)
+		})
+	}
+}
+
+func Test_Permission(t *testing.T) {
+	t.Parallel()
+
+	// New config files should have 0600 permissions
+	t.Run("new config file", func(t *testing.T) {
+		t.Parallel()
+		assert := assert.New(t)
+
+		dir, err := os.MkdirTemp("", "rancher-cli-test-*")
+		assert.NoError(err)
+		defer os.RemoveAll(dir)
+
+		path := filepath.Join(dir, "cli2.json")
+		conf, err := LoadFromPath(path)
+		assert.NoError(err)
+
+		err = conf.Write()
+		assert.NoError(err)
+
+		info, err := os.Stat(path)
+		assert.NoError(err)
+		assert.Equal(os.FileMode(0600), info.Mode())
+
+		// make sure new file doesn't create permission warnings
+		warnings, err := GetFilePermissionWarnings(path)
+		assert.NoError(err)
+		assert.Len(warnings, 0)
+	})
+	// Already existing config files should keep their current permissions
+	t.Run("existing config file", func(t *testing.T) {
+		t.Parallel()
+		assert := assert.New(t)
+
+		dir, err := os.MkdirTemp("", "rancher-cli-test-*")
+		assert.NoError(err)
+		defer os.RemoveAll(dir)
+
+		path := filepath.Join(dir, "cli2.json")
+		err = os.WriteFile(path, []byte(validFile), 0700)
+		assert.NoError(err)
+
+		conf, err := LoadFromPath(path)
+		assert.NoError(err)
+
+		err = conf.Write()
+		assert.NoError(err)
+
+		info, err := os.Stat(path)
+		assert.NoError(err)
+		assert.Equal(os.FileMode(0700), info.Mode())
+	})
+}
+
 func Test_LoadFromPath(t *testing.T) {
 	t.Parallel()
 
--- a/config/config_test.go
+++ b/config/config_test.go
@@ -1,0 +1,108 @@
+package config
+
+import (
+	"os"
+	"path/filepath"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+const (
+	validFile = `
+{
+  "Servers": {
+    "rancherDefault": {
+      "accessKey": "the-access-key",
+      "secretKey": "the-secret-key",
+      "tokenKey": "the-token-key",
+      "url": "https://example.com",
+      "project": "cluster-id:project-id",
+      "cacert": "",
+      "kubeCredentials": null,
+      "kubeConfigs": null
+    }
+  },
+  "CurrentServer": "rancherDefault"
+}`
+	invalidFile = `invalid config file`
+)
+
+func Test_LoadFromPath(t *testing.T) {
+	t.Parallel()
+
+	tests := []struct {
+		name         string
+		content      string
+		expectedConf Config
+		expectedErr  bool
+	}{
+		{
+			name:    "valid config",
+			content: validFile,
+			expectedConf: Config{
+				Servers: map[string]*ServerConfig{
+					"rancherDefault": {
+						AccessKey: "the-access-key",
+						SecretKey: "the-secret-key",
+						TokenKey:  "the-token-key",
+						URL:       "https://example.com",
+						Project:   "cluster-id:project-id",
+						CACerts:   "",
+					},
+				},
+				CurrentServer: "rancherDefault",
+			},
+		},
+		{
+			name:    "invalid config",
+			content: invalidFile,
+			expectedConf: Config{
+				Servers: map[string]*ServerConfig{},
+			},
+			expectedErr: true,
+		},
+		{
+			name:    "non existing file",
+			content: "",
+			expectedConf: Config{
+				Servers:       map[string]*ServerConfig{},
+				CurrentServer: "",
+			},
+		},
+	}
+
+	for _, tt := range tests {
+		tt := tt
+		t.Run(tt.name, func(t *testing.T) {
+			t.Parallel()
+			assert := assert.New(t)
+
+			dir, err := os.MkdirTemp("", "rancher-cli-test-*")
+			assert.NoError(err)
+			defer os.RemoveAll(dir)
+
+			path := filepath.Join(dir, "cli2.json")
+			// make sure the path points to the temp dir created in the test
+			tt.expectedConf.Path = path
+
+			if tt.content != "" {
+				err = os.WriteFile(path, []byte(tt.content), 0600)
+				assert.NoError(err)
+			}
+
+			conf, err := LoadFromPath(path)
+			if tt.expectedErr {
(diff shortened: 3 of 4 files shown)


config/config_test.go
```diff
@@ -0,0 +1,56 @@
+package config
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func Test_LoadFromPath(t *testing.T) {
```
MbolotSuse, line 9: I think it would also be good to have a test that verifies that even when the file doesn't exist it returns a valid config.

config/config_test.go
MbolotSuse, line 1: Given that one of the purposes of this PR is to ensure that future files are created with `0600` permissions, I think it would be a good idea to create the following tests for write:
-  No file exists - new file is created with 0600 permissions
- A file exists with non-0600 permissions - we write the content, but the permissions remain the same

config/testdata/bad-cli2.json
MbolotSuse, line 1: Rather than committing these files, you can include test setup code that creates these files, runs the expected tests, and removes the files. This will allow you to run the tests without introducing hard-coded test files into the repo. I would recommend taking this approach.

MbolotSuse (changes requested):
Couple of small changes on:
- More tests (mostly for write, one new one for load as well)
- Changing test data to not be committed to our repo
- Only calling warn on load and not on write as well

config/config.go
```diff
@@ -41,15 +67,44 @@ func (c Config) Write() error {
...
+	}
+
 	return json.NewEncoder(output).Encode(c)
 }
 
+var warned bool
+
+// The configuration file may have sensitive information (eg: creds) so
+// warn against possibly insecure permissions.
+func warnInsecurePermissions(path string) error {
```
maxsokolovsky, line 87: I'd avoid this global state with the `warned` variable and just return a warning as a formatted string. Then you can log the warning if needed. It's easier to test and use this way.

config/config.go
```diff
@@ -32,6 +33,31 @@ type ServerConfig struct {
...
 
+func LoadFromPath(path string) (Config, error) {
+	cf := Config{
+		Path:    path,
+		Servers: make(map[string]*ServerConfig),
+	}
+
+	content, err := os.ReadFile(path)
+	if os.IsNotExist(err) {
+		return cf, nil
```
maxsokolovsky, line 44: What happens next in this case? Will the be an error down the line if the file doesn't exist?

maxsokolovsky (changes requested):

config/testdata/bad-cli2.json
tomleb: What benefit do you see in generating the test files in that case? Are you suggesting having test setup code that generates the files with random values, or having the content of the test file in the code like this?

```go
var theFile = `{foobar}`
```

tomleb: Ok, I have used `testdata` dirs in the past quite a bit, but I don't have a strong opinion on this so I'll make the suggested change.

config/config_test.go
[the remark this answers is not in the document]
tomleb: Since this comment and the one about `warned` depends on another, can you give me your opinion here: https://github.com/rancher/cli/pull/341#discussion_r1330444926?

config/config.go
tomleb: If the file doesn't exist we return a valid config that only has the `Path` and an initialized map. I'll add a comment to reduce confusion about this.

maxsokolovsky: Should we return an error instead?

config/config_test.go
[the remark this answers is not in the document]
maxsokolovsky: Even if you deliberately ran the tests sequentially, that global state with `warned` is not really needed and thus best to avoid.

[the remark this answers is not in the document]
tomleb: I understand that it's best to be avoided, but the reason I put `warned` is explained in the comment I linked above. Basically, I have chosen to warn only once during the invocation of the CLI instead of everytime the config is read. This is what `warned` does. In light of that, do you have a suggestion? (I'd suggest replying in the comment linked so it's easier to follow)

config/config.go
```diff
@@ -41,15 +67,44 @@ func (c Config) Write() error {
...
 	p := c.Path
 	c.Path = ""
-	output, err := os.Create(p)
+	output, err := os.OpenFile(p, os.O_RDWR|os.O_CREATE|os.O_TRUNC, 0600)
 	if err != nil {
 		return err
 	}
 	defer output.Close()
 
+	if err := warnInsecurePermissions(p); err != nil {
```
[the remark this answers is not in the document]
maxsokolovsky: I believe the Helm CLI issues a warning every time it needs to read the Kubeconfig. Perhaps it's not terrible to do the same here.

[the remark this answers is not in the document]
tomleb: For example, here's what I see if I remove this warned stuff:

```
$ ./bin/rancher kubectl
WARN[0000] Rancher configuration file /home/tlebreux/.rancher/cli2.json is group-readable. This is insecure. 
WARN[0000] Rancher configuration file /home/tlebreux/.rancher/cli2.json is world-readable. This is insecure. 
WARN[0001] Rancher configuration file /home/tlebreux/.rancher/cli2.json is group-readable. This is insecure. 
WARN[0001] Rancher configuration file /home/tlebreux/.rancher/cli2.json is world-readable. This is insecure. 
WARN[0001] Rancher configuration file /home/tlebreux/.rancher/cli2.json is group-readable. This is insecure. 
WARN[0001] Rancher configuration file /home/tlebreux/.rancher/cli2.json is world-readable. This is insecure. 
kubectl controls the Kubernetes cluster manager.
...
```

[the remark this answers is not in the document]
maxsokolovsky: Ah, then, yes, for sure! We should definitely warn only once. It's OK then to run the tests sequentially. But I still think the printing/logging needs to be separate from the function that inspects file stat and returns a string warning.

config/config.go
tomleb: So this is used pretty much everywhere in the CLI and I think I'd like to restrict the scope of the PR a bit. Since it was design this way and was working fine, would you be okay if I kept the same behavior but instead change the name to something else like:
`LoadOrNew()`, `GetOrCreate()` ? This way the name is more clear that the function will either load the config, or return an error, or return a new config?

MbolotSuse: @maxsokolovsky I don't agree with the suggestion that this should return an error. The file not existing is a valid condition - the cli needs to be able to handle this and create the config file during this usage.

I think the code works as-is, would not recommend any changes here.

maxsokolovsky: Oh, so if it doesn't exist, then create it? Makes sense...

MbolotSuse: Yeah that's a valid condition for the CLI as a whole, I think in this case it is expecting to always write to the file so there's no reason to create it here, though @tomleb can correct me if  I am wrong.

maxsokolovsky: Given the old name, which implies it loads the config, It's still not great because then the function has additional, unintuitive behavior. Hence Tom's suggestion to name it something else.

tomleb: Here are some examples without a config file:

```
$ ./bin/rancher server current
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x30 pc=0xa14689]

...

$ ./bin/rancher server ls
FATA[0000] no servers are currently configured 

$ ./bin/rancher kubectl
FATA[0000] no configuration found, run `login` 
```

So not all commands writes a config, and some panics. This isn't good, but I believe should be handled separately from this PR.

config/config.go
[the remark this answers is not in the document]
tomleb: Discussed in Slack, the new approach is to have a public func in `config/` to check the permission and that's used in the app.Before() call. This way, we run this warning only once. One gotcha with this is that we could see the warning appear even if the CLI doesn't read the config but we agreed it's unlikely and okay for now.

cmd/common.go
```diff
@@ -259,27 +258,15 @@ func verifyCert(caCert []byte) (string, error) {
...
-	content, err := ioutil.ReadFile(path)
-	if os.IsNotExist(err) {
-		return cf, nil
-	} else if err != nil {
-		return cf, err
-	}
-
-	err = json.Unmarshal(content, &cf)
-	cf.Path = path
+	return filepath.Join(path, cfgFile)
```
tomleb, line 264: It might even be a good idea to move `cfgFile` to `config.go` and then `LoadFromPath` would accept a directory instead. But I think it's okay as well like this.

config/config_test.go
tomleb: I added the tests as suggested. I have also added more tests to the new `CheckPermission` function to make sure the appropriate number of warning is emitted.

config/config.go
tomleb: The `warned` global variable has been removed as suggested.

config/config_test.go
tomleb: I have added this test case as suggested :+1:

tomleb:
Made subtests parallel as suggested.

config/config.go
```diff
@@ -32,6 +33,57 @@ type ServerConfig struct {
...
+	cf.Path = path
+
+	return cf, nil
+}
+
+// CheckPermission returns the following warnings based on the file permission:
+// - one warning if the file is group-readable
+// - one warning if the file is world-readable
+// We want this because configuration may have sensitive information (eg: creds).
+// No error is returned if the file doesn't exists.
```
maxsokolovsky, line 66:
```suggestion
// A nil error is returned if the file doesn't exist.
```

config/config.go
```diff
@@ -32,6 +33,57 @@ type ServerConfig struct {
...
+func CheckPermission(path string) ([]string, error) {
+	info, err := os.Stat(path)
+	if err != nil {
+		if os.IsNotExist(err) {
+			return []string{}, nil
+		}
+		return []string{}, fmt.Errorf("get file info: %w", err)
+	}
+
+	warnings := []string{}
```
maxsokolovsky, line 76: Nit: prefer explicit zero values
```suggestion
	var warnings []string
```

config/config_test.go
```diff
@@ -0,0 +1,205 @@
...
+
+import (
+	"os"
+	"path/filepath"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+var (
```
maxsokolovsky, line 11: Nit: can be constants.

config/config_test.go
```diff
@@ -0,0 +1,205 @@
...
+		t.Run(tt.name, func(t *testing.T) {
+			t.Parallel()
+			assert := assert.New(t)
+
+			dir, err := os.MkdirTemp("", "rancher-cli-test-*")
+			assert.NoError(err)
+			defer os.RemoveAll(dir)
+
+			path := filepath.Join(dir, "cli2.json")
+			os.WriteFile(path, []byte(validFile), tt.mode)
```
maxsokolovsky, line 71: You're not really checking the error:
```suggestion
			err = os.WriteFile(path, []byte(validFile), tt.mode)
			assert.NoError(err)
```

config/config_test.go
```diff
@@ -0,0 +1,205 @@
...
+		assert.NoError(err)
+
+		conf.Write()
+		assert.NoError(err)
+
+		info, err := os.Stat(path)
+		assert.NoError(err)
+		assert.Equal(os.FileMode(0600), info.Mode())
+
+		// make sure new file don't create permission warnings
```
maxsokolovsky, line 104:
```suggestion
		// make sure new file didn't create permission warnings
```

config/config_test.go
```diff
@@ -0,0 +1,205 @@
...
+
+		dir, err := os.MkdirTemp("", "rancher-cli-test-*")
+		assert.NoError(err)
+		defer os.RemoveAll(dir)
+
+		path := filepath.Join(dir, "cli2.json")
+		conf, err := LoadFromPath(path)
+		assert.NoError(err)
+
+		conf.Write()
```
maxsokolovsky, line 97: You're not checking the error here:
```suggestion
		err = conf.Write()
		assert.NoError(err)
```

config/config_test.go
```diff
@@ -0,0 +1,205 @@
...
+	t.Run("existing config file", func(t *testing.T) {
+		t.Parallel()
+		assert := assert.New(t)
+
+		dir, err := os.MkdirTemp("", "rancher-cli-test-*")
+		assert.NoError(err)
+		defer os.RemoveAll(dir)
+
+		path := filepath.Join(dir, "cli2.json")
+		os.WriteFile(path, []byte(validFile), 0700)
```
maxsokolovsky, line 119: You're not checking the error here:
```suggestion
		err = os.WriteFile(path, []byte(validFile), 0700)
		assert.NoError(err)
```

config/config_test.go
```diff
@@ -0,0 +1,205 @@
...
+		tt := tt
+		t.Run(tt.name, func(t *testing.T) {
+			t.Parallel()
+			assert := assert.New(t)
+
+			dir, err := os.MkdirTemp("", "rancher-cli-test-*")
+			assert.NoError(err)
+			defer os.RemoveAll(dir)
+
+			path := filepath.Join(dir, "cli2.json")
```
maxsokolovsky, line 70: Aren't you writing to the same `tempDir/cli2.json` file concurrently, since the subtests are run in parallel here?

Nit: completely optional, but you could make a temp file as well with `os.CreateTemp()` in the temp dir. Then you could run the subtests in parallel. I believe you could do the same with the other tests.

config/config.go
```diff
@@ -32,6 +33,57 @@ type ServerConfig struct {
...
+			return cf, nil
+		}
+
+		return cf, err
+	}
+
+	if err := json.Unmarshal(content, &cf); err != nil {
+		return cf, fmt.Errorf("unmarshaling %s: %w", path, err)
+	}
+	cf.Path = path
```
maxsokolovsky, line 57: I don't think this is needed, since you do it above already.

config/config.go
```diff
@@ -32,6 +33,57 @@ type ServerConfig struct {
...
+
+	return cf, nil
+}
+
+// CheckPermission returns the following warnings based on the file permission:
+// - one warning if the file is group-readable
+// - one warning if the file is world-readable
+// We want this because configuration may have sensitive information (eg: creds).
+// No error is returned if the file doesn't exists.
+func CheckPermission(path string) ([]string, error) {
```
maxsokolovsky, line 67: Nit: perhaps we may want a better name here. `CheckPermission` doesn't tell me much about whose permission(s) we're checking given a string `path`, and why it returns a slice of strings.
Maybe something like `FilePermissionWarnings`?

cmd/common.go
maxsokolovsky: Agreed! In fact, a few of these functions in this dumping ground (the `common` package) could be moved to the `config` package. `GetConfigPath`, `loadConfig` are the first among them, and the constant as well. But this refactoring is likely outside the scope of this PR.

config/config.go
```diff
@@ -32,6 +33,57 @@ type ServerConfig struct {
 	KubeConfigs     map[string]*api.Config     `json:"kubeConfigs"`
 }
 
+// LoadFromPath attempts to load a config from the given path. If the config
```
maxsokolovsky, line 36:
```suggestion
// LoadFromPath attempts to load a config from the given file path. If the file
```

maxsokolovsky (changes requested):

config/config_test.go
tomleb: Good find!

config/config.go
tomleb: The only concern here is whether `path` was ever written to the config file. Looking at the git history, it seems like it was always `omitempty` so we can probably assume that `path` was never written. In that case, it might be okay to remove that line.. In fact, I think `Config` should be split into 2 separate struct, one with **only** valid config file fields and the other with more stuff like path. The question is how much change do we want to make here? Should we put more time into this config module now, or later?

@MbolotSuse Can you confirm when we want this in? Do we want it for bug complete 2.8 (which is today) or should we take more time to flesh this out a bit more?

config/config.go
tomleb: Changed it to `GetFilePermissionWarnings`, I'll assume you're fine with it and resolve it.

config/config_test.go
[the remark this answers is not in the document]
tomleb: All tests are running in parallel now.

main.go
```diff
@@ -70,6 +71,16 @@ func mainErr() error {
 		if ctx.GlobalBool("debug") {
 			logrus.SetLevel(logrus.DebugLevel)
 		}
+
+		path := cmd.GetConfigPath(ctx)
+		warnings, err := config.GetFilePermissionWarnings(path)
+		if err != nil {
+			return err
```
MbolotSuse, line 78: I don't think that it make sense to return an error in this case. I'd be concerned that a failure here would prevent the rest of the application from running, and since we don't use this call for anything except warning, I don't think that this should be a blocking failure.

config/config.go
tomleb: I'm thinking something like:

```go
// Already hardcoded currently in cmd/common.go
const cfgFile = "cli2.json"
type Environment struct {
	dir string
}

func NewEnvironment() *Environment {}
// LoadConfig gets the config (<dir>/cli2.json) or creates it.
func (e *Environment) LoadConfig() (Config, error) {} 
// SaveConfig saves the config to <dir>/cli2.json
func (e *Environment) SaveConfig(Config) error {}

type Config struct {
	Servers map[string]*ServerConfig
	CurrentServer string
}
```

so 
- Config no longer has `Path` which wasn't written to the config file anyway
- We can move cfgFile from `cmd/common.go` to `config/config.go`. It's fine to be hardcoded because the user can't change it anyway (the `--config` flag specifies the directory, not the file)
- We can remove those hacky Path reset stuff in the tests

config/config_test.go
```diff
@@ -0,0 +1,205 @@
...
+		})
+	}
+}
+
+func Test_Permission(t *testing.T) {
+	t.Parallel()
+
+	// New config files should have 0600 permissions
+	t.Run("new config file", func(t *testing.T) {
+		t.Parallel()
```
MbolotSuse, line 86: I don't think that this one and the `existing config file` can be marked as parallel since they use the same underlying file name. While you can keep the Parallel marker for the permission function (on l82), I don't think that this one (and the one in L111) is safe.

config/config_test.go
```diff
@@ -0,0 +1,205 @@
...
+						CACerts:   "",
+					},
+				},
+				CurrentServer: "rancherDefault",
+			},
+		},
+		{
+			name:        "invalid config",
+			content:     invalidFile,
+			expectedErr: true,
```
MbolotSuse, line 163: Nit: This function has a bit of an odd output. In the cases of an error, it still returns the default config file. Since this function is exported, I think that this should be something that we assert on, since the caller could choose to log the error and do some other processing with the default value returned alongside the error.

MbolotSuse (commented):
Couple of small items, but mostly looks good.

config/config_test.go
tomleb: The `cli2.json` file is created in different folders since each subtest calls `MkdirTemp` so I believe it's safe. Wdyt?

main.go
tomleb: What would make more sense between ignoring the error completely vs logging the error and continuing?

config/config_test.go
tomleb: I might be misunderstanding so apologies in advance.

I believe if `LoadFromPath` returns an error, the config return shouldn't be used, so just good old idiomatic go error handling. However, I see that currently we're not returning an empty config, we're returning a default valid one with the path set.

I think it would make sense to make `LoadFromPath` return an empty config on error, that empty config shouldn't be used, so we don't really need to assert it.

So:
- File already exists: return the config, nil
- File doesn't exist: return default config, nil
- Error reading file: return zero value config, err

Let me know.

tomleb: I have made the changed mentioned above, so I think we don't need to add an assert.

config/config.go
MbolotSuse: The change that you are talking about here seems more significant. I would say that we should try and keep the blast radius on this change small.

tomleb: Agreed.

main.go
MbolotSuse: Agree with your choice, I think logging is the most consistent choice.

MbolotSuse (approved):

JonCrowther (approved):

PR #343, opened 2023-09-28
[v2.8] Update saml auth login endpoint to point toward dashboard endpoint

eliyamlevy:
Parent Issue: https://github.com/rancher/rancher/issues/41651
As per changes in https://github.com/rancher/dashboard/pull/9693 I have updated the SAML Auth login endpoint to use the new dashboard login path.


crobby (approved):
lgtm.  The url lines-up with the dashboard ticket.

jorhett:
This new URL causes the browser to never return, so the terminal hangs... and the token is never be returned. I assume this is a difference on the server side?

PR #346, opened 2024-02-01
Azuread support

enrichman:
## Ref:
- https://github.com/rancher/rancher/issues/42939

SURE-4275

---

This PR adds the support for the AzureAD provider. When configured the `rancher token` command will prompt the user to select it, or use the local login.

The CLI will use the Device Authorization Grant, so it will ask the user to enter a code in the URL, as discussed in the RFC.

The PR adds also some refinements and refactor:

The `getAuthProviders` func now returns a `[]TypedProvider`. With https://github.com/rancher/rancher/pull/44285 the `GetType` method was added, so that every provider automatically implements any interface with the same method.

```go
type TypedProvider interface {
	GetType() string
}

func getAuthProviders(server string) ([]TypedProvider, error) {
	// fetch providers
}
```

With this we can actually do a type assertion on the returned interface and get the concrete underlying type.

Note: before merging the import for the struct will need to be updated.


crobby (commented):
One thought I had about the --prompt option that I figured I'd capture here before I forget.  While it certainly works well, it feels like a bit of a non-standard approach to this sort of operation (copy/paste a URL from a blank page).  We should definitely make sure that this UX is reasonable, and if it is, we should definitely add some extra text in the cli output to make it very clear to the user what the expected flow is in this scenario.

enrichman:
> One thought I had about the --prompt option that I figured I'd capture here before I forget. While it certainly works well, it feels like a bit of a non-standard approach to this sort of operation (copy/paste a URL from a blank page). We should definitely make sure that this UX is reasonable, and if it is, we should definitely add some extra text in the cli output to make it very clear to the user what the expected flow is in this scenario.

@crobby that's a legit concern. What about having a Rancher UI page that is not doing anything in the backend, but just showing you the URL, or maybe the query params? That will be pretty cool, and actually useful also to show the errors. The user will be prompted to input the code, and/or the other parameters.

It need a bit of work on the UI side, but not much.

enrichman:
Just to have something visual this is what I was trying to come up:

Screenshot from 2024-02-12 17-38-44

Please don't pay too much attention to the graphic, I'm not a frontend guy. It will be nice to have also a "copy" button, so the user can easily copy-paste the code, or whatever he needs.

enrichman:
In the final implementation, implementing the RFC there is no update on the UI.

cmd/kubectl_token.go
```diff
@@ -342,12 +362,12 @@ func loginAndGenerateCred(input *LoginInput) (*config.ExecCredential, error) {
 
 func basicAuth(input *LoginInput, tlsConfig *tls.Config) (managementClient.Token, error) {
 	token := managementClient.Token{}
-	username, err := customPrompt("username", true)
+	username, err := customPrompt("Enter username: ", true)
```
crobby, line 365: Thanks for making the prompt look nicer!

crobby (commented):
I think this looks pretty good.
Any chance of adding some unit tests?  Might be reasonable to add a few at least for the new functionality.

enrichman:
Thanks @crobby. I've added just a couple of tests to the `getAuthProviders` func. Others are probably possible, but not without a bit of a refactors.

With the latest commit I've added a `getClient` func, used to create the HTTP client once, with the same TLS config. Otherwise the `skipVerify` or the `caCerts` options were going to be ignored in some requests. I guess this was the case for the BasicAuth, so there was probably a bug.

If that's the case I could try to reproduce it, open the bug, fix it in an isolated way and keep the two PRs smaller.

/cc FYI @samjustus @anupama2501

cmd/kubectl_token.go
```diff
@@ -579,56 +627,73 @@ func generateKey() (string, error) {
...
-		config.RootCAs = roots
+	// load custom certs
+	cert, err := loadAndVerifyCert(caCerts)
+	if err != nil {
+		return nil, err
+	}
+
+	// enrichman: should we load the system CertPool instead of creating a new one?
+	roots := x509.NewCertPool()
+	ok := roots.AppendCertsFromPEM([]byte(cert))
```
andreas-kupries, line 660: Epinio tried to get and extend the system cert pool first, and then fell back to extending a new empty cert pool

cmd/kubectl_token_test.go
```diff
@@ -0,0 +1,114 @@
...
+	}{
+		{
+			name:   "response ok",
+			server: setupServer(responseOK),
+			expectedProviders: []TypedProvider{
+				&apiv3.AzureADProvider{
+					AuthProvider: apiv3.AuthProvider{
+						Type: "azureADProvider",
+					},
+					RedirectURL: "https://login.microsoftonline.com/258928db-3ed6-49fb-9a7e-52e492ffb066/oauth2/v2.0/authorize?client_id=56168f69-a732-48e2-aa21-8aa0909d0976&redirect_uri=https://rancher-local.enrichman.it/verify-auth-azure&response_type=code&scope=openid",
```
andreas-kupries, line 35: Will the tests fail if the `...enrichman.it` domain does not exist ?
If yes, how can we make this test independent of your machines ?
If not, could a placeholder be used which does not refer to your machines ?

cmd/kubectl_token_test.go
```diff
@@ -0,0 +1,114 @@
...
+            "authUrl": "https://login.microsoftonline.com/258928db-3ed6-49fb-9a7e-52e492ffb066/oauth2/v2.0/authorize",
+            "baseType": "authProvider",
+            "clientId": "56168f69-a732-48e2-aa21-8aa0909d0976",
+            "creatorId": null,
+            "deviceAuthUrl": "https://login.microsoftonline.com/258928db-3ed6-49fb-9a7e-52e492ffb066/oauth2/v2.0/devicecode",
+            "id": "azuread",
+            "links": {
+                "self": "…/v3-public/azureADProviders/azuread"
+            },
+            "redirectUrl": "https://login.microsoftonline.com/258928db-3ed6-49fb-9a7e-52e492ffb066/oauth2/v2.0/authorize?client_id=56168f69-a732-48e2-aa21-8aa0909d0976&redirect_uri=https://rancher-local.enrichman.it/verify-auth-azure&response_type=code&scope=openid",
```
andreas-kupries, line 91: See above, references to personal machines

andreas-kupries (changes requested):

cmd/kubectl_token.go
enrichman: I agree. I would like a confirmation from @pjbgf because I'm not sure if I misinterpreted a comment from him.

cmd/kubectl_token_test.go
enrichman: Hm, no, it will not fail because it's just a static check, but still it's better to avoid references to local machine. It's better to have something more generic. :+1:

cmd/kubectl_token_test.go
enrichman: Done with f50f954

cmd/kubectl_token_test.go
enrichman: Done with f50f954

andreas-kupries (dismissed):
The main issue with the local references is done.
Regarding the cert pool I defer to @pjbgf 's final assessment

crobby (commented):
I think this lgtm pending the feedback/resolution on the cert pool.

enrichman:
As discussed internally we can keep the changes as-is (reverting a func name) to keep the existing behaviour. We will address the issues in a next version of the cli.

/cc @crobby @andreas-kupries

pjbgf (approved):

andreas-kupries (approved):

PR #347, opened 2024-02-07
Cleanup of `rancher server` commands

enrichman:
Superseded by:
- https://github.com/rancher/cli/pull/348
- https://github.com/rancher/cli/pull/349

---
old description
---

Fix https://github.com/rancher/rancher/issues/44332

# Bug Description

Some workflows will trigger panics due to assignment to nil maps, or a missing check:

```go
// sc.KubeCredentials is nil -> panics
sc.KubeCredentials[id] = cred
```

```go
serverName := cf.CurrentServer
// missing check for the existance of the value -> panics
URL := cf.Servers[serverName].URL
```

To reproduce just run a `rancher server current` without any configuration:

```
-> % rancher server current 
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x30 pc=0xa14869]

goroutine 1 [running]:
github.com/rancher/cli/cmd.serverCurrent(0xc000195b80?)
        /go/src/github.com/rancher/cli/cmd/server.go:74 +0xc9
github.com/urfave/cli.HandleAction({0xaa5b00?, 0xc28200?}, 0x7?)
...
```

or run a `rancher login` followed by a `rancher token`:

```
-> % rancher login https://myrancher.com --name https://myrancher.com --token MYTOKEN
INFO[0000] Saving config to /home/enrico/.rancher/cli2.json

###

-> % rancher token --server https://myrancher.com --user local
https://myrancher.com/v3-public/authProviders 
Enter credentials for localProvider 
 
Enter username: admin 
Enter password: 
panic: assignment to entry in nil map

goroutine 1 [running]:
github.com/rancher/cli/cmd.cacheCredential(0xc000274160, 0xc00032a2d0, {0xc0003326a0, 0x6})
        /go/src/github.com/rancher/cli/cmd/kubectl_token.go:283 +0x11a
github.com/rancher/cli/cmd.runCredential(0xc000274160)
        /go/src/github.com/rancher/cli/cmd/kubectl_token.go:165 +0x4a5
github.com/urfave/cli.HandleAction({0xaa5b00?, 0xc281f0?}, 0xc00011cfc0?)
        /go/pkg/mod/github.com/urfave/cli@v1.22.5/app.go:524 +0x50
github.com/urfave/cli.(*App).RunAsSubcommand(0xc00011cfc0, 0xc000274000)
        /go/pkg/mod/github.com/urfave/cli@v1.22.5/app.go:410 +0xabd
github.com/urfave/cli.Command.startApp({{0xbd2fb4, 0x5}, {0x0, 0x0}, {0x0, 0x0, 0x0}, {0xbf2956, 0x2e}, {0x0, ...}, ...}, ...)
        /go/pkg/mod/github.com/urfave/cli@v1.22.5/command.go:372 +0xb7f
github.com/urfave/cli.Command.Run({{0xbd2fb4, 0x5}, {0x0, 0x0}, {0x0, 0x0, 0x0}, {0xbf2956, 0x2e}, {0x0, ...}, ...}, ...)
        /go/pkg/mod/github.com/urfave/cli@v1.22.5/command.go:102 +0x845
github.com/urfave/cli.(*App).Run(0xc00011ce00, {0xc00004fd00, 0x6, 0x8})
        /go/pkg/mod/github.com/urfave/cli@v1.22.5/app.go:277 +0xb67
main.mainErr()
        /go/src/github.com/rancher/cli/main.go:136 +0x1f5d
main.main()
        /go/src/github.com/rancher/cli/main.go:57 +0x19
```

# Description

The fix was pretty straightforward (just a couple of check), but I took the opportunity to cleanup a bit the code (only the `server` command, to avoid a huge and "critical" refactor). The server command is pretty safe to play with.

Also some deprecated deps were updated (`terminal` and `ioutil`).

### 1) Common validation

The server commands are expecting a valid server configuration:

```go
func validateServersConfig(cnf config.Config) error {
	// do validation validate
}
```

This common validation was changed to a `cli.BeforeFunc` and added in the `Before` of the `server` commands.

### 2) Testing

To make test possible I moved the business logic to their dedicated func, dropping (when possible) the dependency from the `cli.Context`.

I.e.: the `ServerCurrent` now accept an `io.Writer` and a `*config.Config`. Or the `ServerSwitch` now needs only the `*config.Config` and the `serverName`. This made test possible (and tests were added).

### 3) Centralized Config

The `*config.Config` is needed by every commands, and loaded every time in the same (?) way. This can be done only once at the very start of the root command, in the BeforeFunc. Then the Config can be just passed as a dependency to the commands.

Be aware that to make this work a pointer has to be passed downward, and update the underlying struct in the Before. This is because in Go arguments are passe by value, and the Commands are built before the actual Run (obviously). So the pointer will be always the same, but it will point to the just loaded conf.

```go
conf, err := config.LoadFromPath(path)
if err != nil {
	return err
}
*cfg = conf
```

With this approach we can drop all the `loadConfig` in every command:

```go
// not needed anymore
cf, err := loadConfig(ctx)
if err != nil {
	return err
}
```

### 4) Simplifications

Some minor cleanup, like the `getServers(cfg *config.Config) []*serverData` func.
Instead of getting the ordered keys, and then loop through them we can simply use the already existing `serverData` struct.

This will return an ordered slice of `serverData` that will be used by the server prompt selection, and also by the `ServerLs`, that was using an unordered set of servers (now it's consistent).

### DIFF
--- a/cmd/kubectl_token.go
+++ b/cmd/kubectl_token.go
@@ -11,7 +11,6 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
-	"io/ioutil"
 	"math/big"
 	"net/http"
 	url2 "net/url"
@@ -24,7 +23,7 @@ import (
 	"github.com/rancher/norman/types/convert"
 	managementClient "github.com/rancher/rancher/pkg/client/generated/management/v3"
 	"github.com/urfave/cli"
-	"golang.org/x/crypto/ssh/terminal"
+	"golang.org/x/term"
 )
 
 const deleteExample = `
@@ -109,7 +108,7 @@ func CredentialCommand() cli.Command {
 			},
 		},
 		Subcommands: []cli.Command{
-			cli.Command{
+			{
 				Name:   "delete",
 				Usage:  deleteCommandUsage,
 				Action: deleteCachedCredential,
@@ -440,7 +439,7 @@ func samlAuth(input *LoginInput, tlsConfig *tls.Config) (managementClient.Token,
 			if err != nil {
 				return token, err
 			}
-			content, err := ioutil.ReadAll(res.Body)
+			content, err := io.ReadAll(res.Body)
 			if err != nil {
 				res.Body.Close()
 				return token, err
@@ -606,7 +605,7 @@ func request(method, url string, body io.Reader) ([]byte, error) {
 		return nil, err
 	}
 	defer res.Body.Close()
-	response, err = ioutil.ReadAll(res.Body)
+	response, err = io.ReadAll(res.Body)
 	if err != nil {
 		return nil, err
 	}
@@ -619,7 +618,7 @@ func customPrompt(field string, show bool) (result string, err error) {
 		_, err = fmt.Fscan(os.Stdin, &result)
 	} else {
 		var data []byte
-		data, err = terminal.ReadPassword(int(os.Stdin.Fd()))
+		data, err = term.ReadPassword(int(os.Stdin.Fd()))
 		result = string(data)
 		fmt.Fprintf(os.Stderr, "\n")
 	}
--- a/cmd/server.go
+++ b/cmd/server.go
@@ -30,12 +30,7 @@ func ServerCommand(cfg *config.Config) cli.Command {
 		Usage: "Operations for the server",
 		Description: `Switch or view the server currently in focus.
 `,
-		Before: func(ctx *cli.Context) error {
-			if len(cfg.Servers) == 0 {
-				return errors.New("no servers are currently configured")
-			}
-			return nil
-		},
+		Before: validateServersConfig(cfg),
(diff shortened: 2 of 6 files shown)


crobby (commented):
Looks good.
One question/nit.  Should we use t.Parallel() for the tests.  Obviously, they are not super time consuming, but probably a nice thing to do.

crobby (commented):
Oh, also.  I like the cleanup, but should we have a separate issue/PR for that?  I could go either way on that one.  The cleanup part doesn't strictly go toward fixing the panic in the issue, but it's definitely an improvement.

enrichman:
Closing in favor of:
 - https://github.com/rancher/cli/pull/348
 - https://github.com/rancher/cli/pull/349

PR #348, opened 2024-02-08
Fix panics in `rancher token` and `rancher server current` commands 

enrichman:
Fix https://github.com/rancher/rancher/issues/44332

As I've moved in this PR only the fix related to the panics. I'm going to then rebase the other PR.

# Bug Description

Some workflows will trigger panics due to assignment to nil maps, or a missing check:

```go
// sc.KubeCredentials is nil -> panics
sc.KubeCredentials[id] = cred
```

```go
serverName := cf.CurrentServer
// missing check for the existance of the value -> panics
URL := cf.Servers[serverName].URL
```

To reproduce just run a `rancher server current` without any configuration:

```
-> % rancher server current 
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x30 pc=0xa14869]

goroutine 1 [running]:
github.com/rancher/cli/cmd.serverCurrent(0xc000195b80?)
        /go/src/github.com/rancher/cli/cmd/server.go:74 +0xc9
github.com/urfave/cli.HandleAction({0xaa5b00?, 0xc28200?}, 0x7?)
...
```

or run a `rancher login` followed by a `rancher token`:

```
-> % rancher login https://myrancher.com --name https://myrancher.com --token MYTOKEN
INFO[0000] Saving config to /home/enrico/.rancher/cli2.json

###

-> % rancher token --server https://myrancher.com --user local
https://myrancher.com/v3-public/authProviders 
Enter credentials for localProvider 
 
Enter username: admin 
Enter password: 
panic: assignment to entry in nil map

goroutine 1 [running]:
github.com/rancher/cli/cmd.cacheCredential(0xc000274160, 0xc00032a2d0, {0xc0003326a0, 0x6})
        /go/src/github.com/rancher/cli/cmd/kubectl_token.go:283 +0x11a
github.com/rancher/cli/cmd.runCredential(0xc000274160)
        /go/src/github.com/rancher/cli/cmd/kubectl_token.go:165 +0x4a5
github.com/urfave/cli.HandleAction({0xaa5b00?, 0xc281f0?}, 0xc00011cfc0?)
        /go/pkg/mod/github.com/urfave/cli@v1.22.5/app.go:524 +0x50
github.com/urfave/cli.(*App).RunAsSubcommand(0xc00011cfc0, 0xc000274000)
        /go/pkg/mod/github.com/urfave/cli@v1.22.5/app.go:410 +0xabd
[...]

PR #349, opened 2024-02-08
Cleanup of `rancher server` command

enrichman:
Fixes https://github.com/rancher/rancher/issues/44374

This PR contains the refactor of the https://github.com/rancher/cli/pull/347.
The panics fix where added with https://github.com/rancher/cli/pull/348

# Description

Fixing the panics I took the opportunity to cleanup a bit the code (only the `server` command, to avoid a huge and "critical" refactor). The server command is pretty safe to play with.

Also some deprecated deps were updated (`terminal` and `ioutil`).

### 1) Common validation

The server commands are expecting a valid server configuration:

```go
func validateServersConfig(cnf config.Config) error {
	// do validation validate
}
```

This common validation was changed to a `cli.BeforeFunc` and added in the `Before` of the `server` commands.

### 2) Testing

To make test possible I moved the business logic to their dedicated func, dropping (when possible) the dependency from the `cli.Context`.

I.e.: the `ServerCurrent` now accept an `io.Writer` and a `*config.Config`. Or the `ServerSwitch` now needs only the `*config.Config` and the `serverName`. This made test possible (and tests were added).

### 3) Centralized Config

The `*config.Config` is needed by every commands, and loaded every time in the same (?) way. This can be done only once at the very start of the root command, in the BeforeFunc. Then the Config can be just passed as a dependency to the commands.

Be aware that to make this work a pointer has to be passed downward, and update the underlying struct in the Before. This is because in Go arguments are passe by value, and the Commands are built before the actual Run (obviously). So the pointer will be always the same, but it will point to the just loaded conf.

```go
conf, err := config.LoadFromPath(path)
if err != nil {
	return err
}
*cfg = conf
```

With this approach we can drop all the `loadConfig` in every command:

```go
// not needed anymore
cf, err := loadConfig(ctx)
if err != nil {
	return err
}
```

### 4) Simplifications

Some minor cleanup, like the `getServers(cfg *config.Config) []*serverData` func.
Instead of getting the ordered keys, and then loop through them we can simply use the already existing `serverData` struct.

This will return an ordered slice of `serverData` that will be used by the server prompt selection, and also by the `ServerLs`, that was using an unordered set of servers (now it's consistent).

### DIFF
--- a/cmd/server.go
+++ b/cmd/server.go
@@ -30,12 +30,7 @@ func ServerCommand(cfg *config.Config) cli.Command {
 		Usage: "Operations for the server",
 		Description: `Switch or view the server currently in focus.
 `,
-		Before: func(ctx *cli.Context) error {
-			if len(cfg.Servers) == 0 {
-				return errors.New("no servers are currently configured")
-			}
-			return nil
-		},
+		Before: validateServersConfig(cfg),
 		Subcommands: []cli.Command{
 			{
 				Name:  "current",
@@ -239,4 +234,13 @@ func getServers(cfg *config.Config) []*serverData {
 	}
 
 	return servers
+}
+
+func validateServersConfig(cfg *config.Config) cli.BeforeFunc {
+	return func(ctx *cli.Context) error {
+		if len(cfg.Servers) == 0 {
+			return errors.New("no servers are currently configured")
+		}
+		return nil
+	}
 }
--- a/cmd/server.go
+++ b/cmd/server.go
@@ -105,6 +105,10 @@ func serverDelete(cfg *config.Config, serverName string) error {
 	}
 	delete(cfg.Servers, serverName)
 
+	if cfg.CurrentServer == serverName {
+		cfg.CurrentServer = ""
+	}
+
 	err := cfg.Write()
 	if err != nil {
 		return err
--- a/cmd/server.go
+++ b/cmd/server.go
@@ -38,7 +38,7 @@ func ServerCommand() cli.Command {
 				Name:  "current",
 				Usage: "Display the current server",
 				Action: func(ctx *cli.Context) error {
-					return ServerCurrent(ctx.App.Writer, cfg)
+					return serverCurrent(ctx.App.Writer, cfg)
 				},
 			},
 			{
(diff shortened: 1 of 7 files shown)


cmd/server_test.go
```diff
@@ -0,0 +1,134 @@
+package cmd_test
```
JonCrowther, line 1: For completeness, could you add tests for `ServerDelete` and `ServerLs`?

JonCrowther (commented):
Overall looks very good, just a couple questions for my understanding.

main.go
```diff
@@ -82,6 +84,12 @@ func mainErr() error {
 			logrus.Warning(warning)
 		}
 
+		conf, err := config.LoadFromPath(path)
```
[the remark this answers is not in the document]
enrichman: Hum, it makes sense. I thought it was a more generic "AppConfig/Context" useful for every command but it's not the case (not yet at least).

Usually this kind of struct contains the global flags used by the subsequent commands, i.e. the `debug` level, or the `output` format. I'm fine on moving it, then we can see if we need to extract something in the future!

cmd/server_test.go
enrichman: Sure thing. I didn't want to clutter too much the PR but I'm happy to add some more test. :smile:

enrichman: Ah, IIRC there was a small issue for testing this. I will need to add a method to the TabWriter, to set and override the Writer.

I think it's something useful anyway, if you want to add the opportunity to set the output to a file, for example.

main.go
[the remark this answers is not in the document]
enrichman: Done with https://github.com/rancher/cli/pull/349/commits/e44ed8a41d9a0b89bddd56286e09006d0545ec22

cmd/server.go
```diff
@@ -42,72 +47,63 @@ func ServerCommand() cli.Command {
...
+						return err
+					}
+					return ServerSwitch(cfg, serverName)
+				},
 			},
 		},
 	}
 }
 
 // serverCurrent command to display the name of the current server in the local config
```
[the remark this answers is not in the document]
enrichman: Done! https://github.com/rancher/cli/pull/349/commits/7d8aff605e56689d9f82a98541526fc30821e94a

cmd/server_test.go
enrichman: Tests for `serverDelete` https://github.com/rancher/cli/pull/349/commits/52441bfdeb783f91c0b5740e315bd044159ea8fd that led to a small fix (`CurrentServer` was kept also after a delete).

And `serverLs` https://github.com/rancher/cli/pull/349/commits/8349461de9f339c709a41a6c3b56db82421e66be

As you can see I had to create a new func to create a TableWriter, to pass just the needed configuration. Now the `io.Writer` can be customized (and used in out tests), also it accept the `quiet` and `format` options.

JonCrowther (approved):
This LGTM. Thank you for adding the changes I suggested!

crobby (approved):
lgtm, happy to see some tests in here!

PR #351, opened 2024-03-14
add username flag for token command

billiford:
This PR allows us to pass the username as a flag to the `rancher token` command so it does not need to be input each time. This is helpful in Kubernetes configs when defining the login command for a given cluster.

Example
```sh
$ ranch token --server=my-server --user=my-user --cluster=my-cluster --auth-provider=activeDirectoryProvider --username my-username
Enter credentials for activeDirectoryProvider

Enter password:
```
As we can see it no longer asks for a username if one is passed in.

### DIFF
--- a/cmd/kubectl_token.go
+++ b/cmd/kubectl_token.go
@@ -42,6 +42,7 @@ Example:
 type LoginInput struct {
 	server       string
 	userID       string
+	username     string
 	clusterID    string
 	authProvider string
 	caCerts      string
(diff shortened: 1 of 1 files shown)


andreas-kupries (commented):
While it looks ok I am wondering about the relation between this new `username` and the pre-existing `userId` (flag `--user`). Is one derivable from the other ?

billiford:
In our environment there is no relation. It is setup so the `userId` is the name of the cluster and is therefore the same for everyone connecting to that cluster. Since we're using an auth provider, the `username` is what's passed to Active Directory to identify the client during login.

andreas-kupries (dismissed):

billiford:
Why not support both? The `--password` flag is useful for my kubectl config login command.

billiford:
I'd be happy to change this to `--authUser` in this PR, but I think anything involving a password should take place in a separate PR.

andreas-kupries (approved):

PR #354, opened 2024-03-19
[v2.8] Make runnable on windows

HarrisonWAffel:
### Issue: https://github.com/rancher/rancher/issues/44844

This is a backport PR of https://github.com/rancher/cli/pull/326 for 2.8.

### DIFF
--- a/cmd/kubectl_token.go
+++ b/cmd/kubectl_token.go
@@ -17,6 +17,7 @@ import (
 	url2 "net/url"
 	"os"
 	"os/signal"
+	"runtime"
 	"strings"
 	"time"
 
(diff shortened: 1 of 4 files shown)


cmd/kubectl_token.go
```diff
@@ -78,6 +77,14 @@ var supportedAuthProviders = map[string]bool{
 }
 
 func CredentialCommand() cli.Command {
+	configDir, err := ConfigDir()
+	if err != nil {
+		if runtime.GOOS == "windows" {
+			configDir = "%HOME%\\.rancher"
```
jakefhyde, line 83: I know this is a backport but is seems to me like this should have `os.ExpandEnv`.

HarrisonWAffel: Yeah thats a good point, though in this case `configDir` is only used in the usage output for the delete subcommand. Depending on the length of `%HOME%` the usage string might get a bit longer than needed

jakefhyde (approved):

PR #357, opened 2024-04-08
Added linter and removed deprecated and unused packages

enrichman:
Ref: 
- https://github.com/rancher/rancher/issues/43185

---

This PR adds a Github Action that perform a lint, test and build to quick check deprecations and errors.

Edited the `golangci-lint` config to run the default linters.
Removed `"io/ioutil"` deprecated package, removed some unused funcs, removed unused `prompt` command (see https://github.com/rancher/cli/pull/115), fixed error checking.

**Removed** direct dependencies:
```
github.com/c-bata/go-prompt v0.2.6
github.com/docker/docker v1.6.1
github.com/gorilla/websocket v1.4.2
github.com/patrickmn/go-cache v2.1.0+incompatible
```

The "new" `golang.org/x/text v0.11.0` dependency was already present as an indirect dep.


cmd/server.go
```diff
@@ -175,7 +175,10 @@ func getSelectedServer(ctx *cli.Context, cfg *config.Config) (string, error) {
 // prompt the user to select one.
 func serverFromInput(ctx *cli.Context, cf *config.Config) (string, error) {
 	servers := getServers(cf)
-	displayListServers(ctx, servers)
+	err := displayListServers(ctx, servers)
```
bigkevmcd, line 178: This could be `if err := displayListServers(ctx, servers); err != nil ... ` ?

cmd/cluster.go
```diff
@@ -747,7 +747,7 @@ func getClusterK8sOptions(c *cliclient.MasterClient) []string {
 	setting, _ := c.ManagementClient.Setting.ByID("k8s-version-to-images")
 	var objmap map[string]*json.RawMessage
 
-	json.Unmarshal([]byte(setting.Value), &objmap)
+	_ = json.Unmarshal([]byte(setting.Value), &objmap)
```
bigkevmcd, line 750: Looking at this function, it looks like there are a couple of cases where it could fail?

What happens if `ByID` returns an error? If the JSON can't be unmarshaled?

I appreciate that this is apparently unlikely (based on the signature of the function) but if we're not going to return the error, maybe we could return an empty slice if `ByID` returns an error?

bigkevmcd (approved):
Thanks for the cleanup!

Neither of the two issues are blockers.

cmd/server.go
enrichman: I'm not a big fan of the one-line err checks, but if it's more idiomatic I'm ok with that: 5d03f2a :smile:

cmd/cluster.go
enrichman: A couple of lines before actually there was an API call without the err check. Because of this it's probably better to check and return an error (it could be helpful to warn the user about some network misconfiguration).

Done with 05fc2fa.

enrichman:
@bigkevmcd thanks for the review! I also found two more unused/old commands.

andreas-kupries (approved):
lgtm

bigkevmcd (approved):

PR #359, opened 2024-04-10
[v2.8] [backport] Added linter and removed deprecated and unused packages (#357)

enrichman:
Ref: 
- https://github.com/rancher/rancher/issues/43185

---

This is a cherrypick of https://github.com/rancher/cli/pull/357 to backport the fixes added to v2.9.

PR #360, opened 2024-04-10
[v2.8] Go1.21 update

enrichman:
Bump of Go to 1.21 to get the latest security fixes, and fix the issue with https://github.com/rancher/cli/pull/359

### DIFF
--- a/Dockerfile.dapper
+++ b/Dockerfile.dapper
@@ -1,8 +1,8 @@
-FROM registry.suse.com/bci/golang:1.20
+FROM registry.suse.com/bci/golang:1.21
 
 RUN zypper -n install docker rsync xz zip
 
-ENV GOLANGCI_LINT v1.53.3
+ENV GOLANGCI_LINT v1.57.1
 RUN curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b $(go env GOPATH)/bin "$GOLANGCI_LINT"
 
 ENV DAPPER_SOURCE /go/src/github.com/rancher/cli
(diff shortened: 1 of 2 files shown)


pmatseykanets:
Q: Do we want to keep Go versions aligned with respective rancher release lines? For now it's still [1.21](https://github.com/rancher/rancher/blob/6e90cf0f9b9b6d6ec3864e082fe949795d3a69fb/go.mod#L3) for v2.8.

pmatseykanets (commented):
Q: Do we want to keep Go versions aligned with respective rancher release lines? For now it's still [1.21](https://github.com/rancher/rancher/blob/6e90cf0f9b9b6d6ec3864e082fe949795d3a69fb/go.mod#L3) for v2.8.

enrichman:
@pmatseykanets that's a good question. I was going to bump to go 1.21 for the same reasoning, and then I decided to go for 1.22 because... why not? I don't have a strong opinion about this, it was just to have more time before stopping getting security updates. :shrug:

crobby:
> @pmatseykanets that's a good question. I was going to bump to go 1.21 for the same reasoning, and then I decided to go for 1.22 because... why not? I don't have a strong opinion about this, it was just to have more time before stopping getting security updates. 🤷

I think my gut instinct would be to use 1.21 here just to be in sync with all of the 2.8 kingdom, but if folks like 1.22 instead, it's probably fine as well.

enrichman:
Let's keep them in sync then! Wait for it. :slightly_smiling_face:

enrichman:
Done! Thanks everyone.

pmatseykanets (approved):

crobby (approved):

PR #329, opened 2022-12-21
Add arm64 build output

SimonyanG:
Support for apple M1 silicon

PR #362, opened 2024-04-19
[v2.9] Migrate from Drone to Github Actions, remove Dapper and add Darwin/arm64 support

enrichman:
This PR migrates the CI to Drone to GitHub Actions. It also removes `dapper`.

Secrets were already migrated and tested with https://github.com/rancher/cli/pull/365.

In order to use the `golangci/golangci-lint-action` the `validate` script was changed to skip the linting, and a dedicated `lint` script was added.

As a bonus the `arm64` arch was added to the Darwin build (`OS_ARCH_ARG[darwin]="amd64 arm64"` ), this will fix #329 .


.github/workflows/release.yml
```diff
@@ -14,18 +14,75 @@ jobs:
...
-      uses: actions/checkout@v3
-
     - name: Login to Docker Hub
       uses: docker/login-action@v3
       with:
         username: ${{ env.DOCKER_USERNAME }}
         password: ${{ env.DOCKER_PASSWORD }}
+
+    # Fail for non-tagged commits
+    - name: Tag Check
```
JonCrowther, line 34: If I understand correctly, this job only triggers on pushing a tag. Do we need this step?

.github/workflows/ci.yml
```diff
@@ -35,15 +67,15 @@ jobs:
       id-token: write # needed for the Vault authentication
 
     steps:
+      - name: Checkout Repo
+        uses: actions/checkout@v3
+
       - name: Load Secrets from Vault
```
raulcabello, line 73: this step is failing https://github.com/rancher/cli/actions/runs/8879832219/job/24378515349?pr=362

.github/workflows/ci.yml
```diff
@@ -9,24 +9,56 @@ jobs:
...
+        echo "VERSION=$VERSION" >> $GITHUB_ENV
+
+    - name: Docker Build
+      uses: docker/build-push-action@v5
+      with:
+        push: false
+        context: package
+        tags: rancher/cli2:${{ env.VERSION }}
 
   fossa:
```
raulcabello, line 63: should we also run fossa in `workflow/release.yaml` ? From what I see in the `drone.yaml` we were also runninfg fossa when creating a tag

.github/workflows/ci.yml
```diff
@@ -9,24 +9,56 @@ jobs:
...
+      if: startsWith(github.ref, 'refs/tags/v')
+      run: echo "GITHUB_TAG=$GITHUB_REF_NAME" >> $GITHUB_ENV
 
     - name: Build
-      run: make build
+      env:
+        CROSS: 1
+      run: ./scripts/build
+
+    - name: Test
```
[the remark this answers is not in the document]
enrichman: :man_facepalming: wrong rebase. Thanks.

.github/workflows/release.yml
enrichman: The job could be also triggered manually with the `workflow_dispatch`, so this check was added to be sure that the build is running against an existing tag.

We can remove it if we want to push "dev" version of the cli.

.github/workflows/ci.yml
enrichman: Yes, that's expected because the action is running on a fork (`Unable to get ACTIONS_ID_TOKEN_REQUEST_URL env variable`).

The merged version is fine: https://github.com/rancher/cli/pull/365#issuecomment-2075210066

I could add a check to see if the action is running on the same repo:

```yaml
if: github.repository_owner == 'rancher'
```

.github/workflows/ci.yml
enrichman: That's fine! I didn't put it because I thought that will be a redundant step.

.github/workflows/ci.yml
raulcabello: wouldn't it be better to just execute it when pushing to a release branch or when creating a tag? We could create a separate workflow, something like we do in kubewarden https://github.com/kubewarden/kubewarden-controller/blob/main/.github/workflows/fossa.yml

raulcabello: wouldn't it be better to execute it just when pushing to a release branch or when creating a tag? We could create a separate workflow for fossa like we do in kubewarden https://github.com/kubewarden/kubewarden-controller/blob/main/.github/workflows/fossa.yml

.github/workflows/release.yml
raulcabello: why do we want to trigger a release manually? In my opinion a release should always be triggered by a tag. I'd remove this tag check and `workflow_dispatch`

.github/workflows/ci.yml
enrichman: Yes, I like it more!

enrichman: Done 3672f7c.

.github/workflows/release.yml
enrichman: No, but in the past I found convenient to have a way to manually trigger a workflow.

Thinking a bit more about this you still have the chance to re-trigger a failed build, so maybe it's not needed. I'll remove it. :)

enrichman: Done eb27b85

JonCrowther (approved):

crobby (approved):
I'm not a GHA expert yet, but nothing seems horribly out of place.  lgtm

raulcabello (approved):

PR #365, opened 2024-04-24
GHA migration: secrets check

enrichman:
This PR adds the steps that will use the migrated secrets,  and a `release.yml` workflow that will be used for the releases.

The FOSSA secret with the API key is used during the CI workflow.
The Docker Hub username and password will be used during the release. At the moment we are just going to login to see that they are working fine.

This PR is needed because the branched of the CLI repo are protected, and we need PRs for them.
This is a first step "soft" PR used for the migration with no side-effects.


.github/workflows/ci.yml
```diff
@@ -25,3 +26,21 @@ jobs:
...
+      contents: read
+      id-token: write # needed for the Vault authentication
+
+    steps:
+      - name: "Load Secrets from Vault"
+        uses: rancher-eio/read-vault-secrets@main
+        with:
+          secrets: |
+            secret/data/github/repo/${{ github.repository }}/fossa/credentials token | FOSSA
+      - uses: actions/checkout@v3
```
andreas-kupries, line 42: I thought for a bit that this belonged to the `load Secrets ...` step instead of being its own.
Needs a bit of separation, and giving it a name would not be bad either.

.github/workflows/release.yml
```diff
@@ -0,0 +1,30 @@
...
+
+    steps:
+    - name: "Load Secrets from Vault"
+      uses: rancher-eio/read-vault-secrets@main
+      with:
+        secrets: |
+          secret/data/github/repo/${{ github.repository }}/dockerhub/rancher/credentials username | DOCKER_USERNAME ;
+          secret/data/github/repo/${{ github.repository }}/dockerhub/rancher/credentials password | DOCKER_PASSWORD ;
+
+    - uses: actions/checkout@v3
```
andreas-kupries, line 24: Name is missing

andreas-kupries (changes requested):

.github/workflows/release.yml
enrichman: I've added a name also to the FOSSA step. :+1:

enrichman:
Thanks for the review @andreas-kupries . I checked with @Anarkis and the error is expected because of the run on a different repo. Once merged we can check that the secrets are ok.
Do you think I should open the PR on a different branch? It's safe to merge on purpose.

andreas-kupries:
> Thanks for the review @andreas-kupries . I checked with @Anarkis and the error is expected because of the run on a different repo. Once merged we can check that the secrets are ok. Do you think I should open the PR on a different branch? It's safe to merge on purpose.

For my migration I actually did a test PR on a non-forked branch to check that the creds worked, and, of course the actions after also. For the FOSSA this might not be as important, as it is a code check, and not pushing anything.

bigkevmcd (approved):
Looks good to me!

enrichman:
Secrets are working fine :+1: 

- FOSSA API key: https://github.com/rancher/cli/actions/runs/8818987612/job/24209107810
- Docker login: https://github.com/rancher/cli/actions/runs/8818997011/job/24209139489

PR #367, opened 2024-05-08
Fix GitHub Action permission to publish assets

enrichman:
The testing release `v2.9.0-rc2` failed: https://github.com/rancher/cli/actions/runs/8999212629/job/24720925994

Error was `HTTP 403: Resource not accessible by integration` because the `contents` permission was set to `read`, while the needed permission to update a release should be `write`.

Added `contents: write` permission in release workflow.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -8,7 +8,7 @@ on:
 jobs:
   release:
     permissions:
-      contents: write
+      contents: write # needed to create/update the release with the assets
       id-token: write # needed for the Vault authentication
     runs-on: ubuntu-latest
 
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -8,7 +8,7 @@ on:
 jobs:
   release:
     permissions:
-      contents: read
+      contents: write
```
bigkevmcd, line 11: Maybe we add a similar comment...

```go
# needed to create the release
```?

bigkevmcd (approved):

.github/workflows/release.yml
enrichman: Yes, added comment!

bigkevmcd (approved):

andreas-kupries (approved):

PR #369, opened 2024-05-23
[v2.9] add codeowners file

JonCrowther:
Add the collie team as the codeowner of rancher/cli

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -1,0 +1,1 @@
+*       @rancher/collie


pmatseykanets (approved):

enrichman (approved):
It shows an error to me:

image

Do we need to make the team public? It seems already the case, but not sure why it fails then.

I'm approving because the @rancher/collie tag seems to work, so... :shrug: 

image

JonCrowther:
@enrichman the error was because the team did not have write access to the repository. The EIO team has fixed that for us and now github is saying the file is valid 😄

PR #383, opened 2024-08-08
enable server ls command to accept the format option

john801205:
Ref: https://github.com/rancher/rancher/issues/48169

### Summary
This PR allows users to specify format for the server ls command.

### Usage

```bash
$ make build
$ ./bin/rancher server ls -h
List all servers

Usage:
  rancher server ls [OPTIONS]

Options:
   --format value  'json', 'yaml' or Custom format: '{{.Name}} {{.URL}}'
$ ./bin/rancher server ls --format '{{.Name}}'
beta
prod
```


crobby (changes requested):
fairly straightforward change, but can you add a unit test for this as well?

john801205:
> fairly straightforward change, but can you add a unit test for this as well?

@crobby
Do you want me to add some unit test for the returned object `cli.Command`?
Otherwise, we already have some unit tests for the main logic here. We just don't allow the input from the command line now.
https://github.com/rancher/cli/blob/f5d2eaabebfc03b4359d333dc82540a8adb09683/cmd/server_test.go#L230-L237

crobby:
> > fairly straightforward change, but can you add a unit test for this as well?
> 
> @crobby Do you want me to add some unit test for the returned configuration object `cli.Command`? Otherwise, we already have some unit tests for the main logic here. We just don't allow the flag from the command line now.
> 
> https://github.com/rancher/cli/blob/f5d2eaabebfc03b4359d333dc82540a8adb09683/cmd/server_test.go#L230-L237

If you can add something like that, I think there is at least some value in it.  I do agree that the "main logic" is indeed already tested here.

john801205:
> If you can add something like that, I think there is at least some value in it. I do agree that the "main logic" is indeed already tested here.

@crobby Sure, just added. Also, I set the functions inside the object to `nil` to skip checking since it's not possible to check if they are equal at this moment.

crobby (dismissed):

andreas-kupries (dismissed):

alegrey91 (dismissed):

cmd/server_test.go
```diff
@@ -7,8 +7,63 @@ import (
 
 	"github.com/rancher/cli/config"
 	"github.com/stretchr/testify/assert"
+	"github.com/urfave/cli"
 )
 
+func TestServerCommand(t *testing.T) {
```
pmatseykanets, line 13: TBH I'm not sure why we need such a test. This is brittle and doesn't really test any functionality. I'd rather not have it at all.
/cc @crobby

crobby: I see your point.  Maybe it would be fine to just remove the test and merge as it was.

pmatseykanets: @john801205 Sorry for the back-and-forth. Can you please revert the unit test changes and we'll merge the PR?

john801205: Sure, no problem. I have reverted the change.

pmatseykanets (approved):

cmd/server_test.go
pmatseykanets: Thank you!

crobby (approved):

PR #183, opened 2018-07-09
Added support for corporate proxy variables.

jomeier:
Hi,

my company has a corporate proxy. rancher-cli runs behind that. Our rancher 2 server instance runs in the public internet.

We want the rancher-cli to connect to the rancher 2 server.

Our corporate proxy prevents rancher-cli from connecting to the internet. The http_proxy, ... variables are ignored.

I added support for environment variables to all calls of 

```
   http.Transport(...)
```

Now it works.

Accept this PR, please.

Thanks a lot and best regards,

Josef

PR #393, opened 2024-09-17
HTTP timeouts and proxy

pgillich:
It's an extended implementation for https://github.com/rancher/cli/pull/183 .


crobby (changes requested):
Is it possible to have some unit tests added to cover this functionality?

pgillich:
> Is it possible to have some unit tests added to cover this functionality?

Yes. Done.

alegrey91:
Hello @pgillich, the PR looks good but I suspect it is a bit over complicated for what we really need.
Can you please explain what's the need of creating the `HTTPClienter` interface and mock the `New()` method?
I might be missing something, so better to ask.

cliclient/cliclient.go
```diff
@@ -25,6 +29,29 @@ type MasterClient struct {
...
+var DefaultHTTPClient HTTPClienter = &HTTPClient{}
+
+/*
+TestingReplaceDefaultHTTPClient replaces DefaultHTTPClient for unit tests.
+Not thread-safe.
+Call the returned function by defer keyword, for example:
+
+	defer cliclient.TestingReplaceDefaultHTTPClient(mockClient)()
+*/
+func TestingReplaceDefaultHTTPClient(mockClient HTTPClienter) func() {
```
alegrey91, line 47: Can we move it undert the `cliclient_test.go` file?

cliclient/cliclient.go
```diff
@@ -99,8 +126,31 @@ func NewProjectClient(config *config.ServerConfig) (*MasterClient, error) {
...
+TestingForceClientInsecure sets testForceClientInsecure to true for unit tests.
+It's a workaround to github.com/rancher/norman/clientbase.NewAPIClient,
+which replaces net/http.Client.Transport (including proxy and TLS config),
+so the client TLS config of net/http/httptest.Server will be lost.
+Not thread-safe.
+Call the returned function by defer keyword, for example:
+
+	defer cliclient.TestingForceClientInsecure()()
+*/
+func TestingForceClientInsecure() func() {
```
alegrey91, line 141: Can we move it undert the `cliclient_test.go` file?

cliclient/cliclient.go
```diff
@@ -99,8 +126,31 @@ func NewProjectClient(config *config.ServerConfig) (*MasterClient, error) {
...
+	origTestForceClientInsecure := testingForceClientInsecure
+	testingForceClientInsecure = true
+	return func() {
+		testingForceClientInsecure = origTestForceClientInsecure
+	}
+}
+
 func (mc *MasterClient) newManagementClient() error {
 	options := createClientOpts(mc.UserConfig)
+	if testingForceClientInsecure {
```
alegrey91, line 151: Personally, I would avoid adding conditions for tests into main code.

cliclient/cliclient_test.go
```diff
@@ -0,0 +1,101 @@
...
+	"time"
+
+	"github.com/stretchr/testify/assert"
+
+	"github.com/rancher/cli/cliclient/mocks"
+	"github.com/rancher/cli/config"
+	"github.com/rancher/norman/types"
+)
+
+func TestCreateClientOpts(t *testing.T) {
```
alegrey91, line 22: Can you move it to `Test_createClientOpts` in order to follow the tests naming convention?

cliclient/cliclient_test.go
```diff
@@ -0,0 +1,101 @@
...
+	assert.NotNil(t, clientOpts)
+	assert.NotNil(t, clientOpts.HTTPClient)
+	assert.NotNil(t, clientOpts.HTTPClient.Transport)
+	assert.Equal(t, "https://a.b/v3", clientOpts.URL)
+	assert.Equal(t, "AccessKey", clientOpts.AccessKey)
+	assert.Equal(t, "SecretKey", clientOpts.SecretKey)
+	assert.Equal(t, "CACerts", clientOpts.CACerts)
+}
+
+func TestDefaultHTTPClient(t *testing.T) {
```
alegrey91, line 41: Can you move it to `TestHTTPClient_New` in order to follow the tests naming convention?

cliclient/cliclient_test.go
```diff
@@ -0,0 +1,101 @@
...
+	transport, is := client.Transport.(*http.Transport)
+	assert.True(t, is)
+	assert.NotNil(t, transport)
+	assert.NotNil(t, transport.DialContext)
+	assert.NotNil(t, transport.Proxy)
+	assert.Equal(t, "net/http.ProxyFromEnvironment", runtime.FuncForPC(reflect.ValueOf(transport.Proxy).Pointer()).Name())
+	assert.Equal(t, 10*time.Second, transport.ResponseHeaderTimeout)
+}
+
+func TestMasterClientNewManagementClient(t *testing.T) {
```
alegrey91, line 57: Can you move it to `TestNewManagementClient` in order to follow the tests naming convention?

pgillich:
> Hello @pgillich, the PR looks good but I suspect it is a bit over complicated for what we actually need.
> Can you please explain what's the need of creating the `HTTPClienter` interface and mock the `New()` method?
> I might be missing something, so better to ask.

Without mocking network communication, below new/changed functions cannot be tested:

- cmd/login.go:getCertFromServer()
- cmd/ssh.go:getSSHKey()
- cliclient/cliclient.go:newManagementClient()

The `HTTPClienter` interface is needed for mocking, because in Go only interfaces can be mocked.

Originally I didn't want to write unit tests, because I could see, the coverage is low and the source code structure is not ready for writing good unit test coverage (for example: missing Go interface). But @crobby asked me to write unit test. So I did it.

cliclient/cliclient.go
pgillich: Yes, we can. But the compile will be failed, because this function is called from another package, too (`cmd/*_test.go`). Go compiler enables importing from `*_test.go files` only from same package.

It's possible to create a new package, for example `testutil`, which can be imported from everywhere. But it may introduce a new claim: a new package. But anyway, if you mind, I can do it.

cliclient/cliclient.go
pgillich: Same situation to `TestingReplaceDefaultHTTPClient`.

cliclient/cliclient.go
pgillich: Me too, but the config.ServerConfig does not support the insecure flag.

If you mind, I can extend the config.ServerConfig with the insecure flag, annotated with `json:"-"`, to make sure, it won't be loaded from config file.

cliclient/cliclient_test.go
pgillich: Done.

cliclient/cliclient_test.go
pgillich: Done

test/ca-cert.pem
```diff
@@ -0,0 +1,21 @@
+-----BEGIN CERTIFICATE-----
```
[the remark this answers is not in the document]
pgillich: Done

cliclient/cliclient.go
alegrey91: I think that creating a new package would be a better approach. What do you think @crobby ?

cliclient/cliclient.go
alegrey91: Can you explain why we need to set the `Insecure` for tests?

alegrey91:
> > Hello @pgillich, the PR looks good but I suspect it is a bit over complicated for what we actually need.
> > Can you please explain what's the need of creating the `HTTPClienter` interface and mock the `New()` method?
> > I might be missing something, so better to ask.
> 
> Without mocking the network communication, below new/changed functions cannot be tested:
> 
> * cmd/login.go:getCertFromServer()
> * cmd/ssh.go:getSSHKey()
> * cliclient/cliclient.go:newManagementClient()
> 
> The `HTTPClienter` interface is needed for mocking, because in Go, only interfaces can be mocked.
> 
> Originally I didn't want to write unit tests, because I could see, the coverage is low and the source code structure is not ready for writing good unit test coverage (for example: missing Go interfaces). But @crobby asked me to write unit test. So I did it.

I got the point and thanks for writing tests. My concern is that we are introducing several things that can be considered superfluous. What I mean is that we can probably avoid of testing the `NewManagementClient()` since we are just checking that the master client is not nil. wdyt @crobby?
If possible I would avoid mocking things when possible.

cliclient/cliclient.go
pgillich: A httptest.Server was started, which supports TLS (client TLS config can be get), but it was unable to use, because github.com/rancher/norman/clientbase.NewAPIClient replaces net/http.Client.Transport (including proxy and TLS config),
+so the client TLS config of net/http/httptest.Server will be lost. I cannot understand why NewAPIClient  was designed this way.

So insecure must be used with httptest.Server.

pgillich:
> I got the point and thanks for writing tests. My concern is that we are introducing several things that can be considered superfluous. What I mean is that we can probably avoid of testing the `NewManagementClient()` since we are just checking that the master client is not nil. wdyt @crobby? If possible I would avoid mocking things when possible.

I understand you would like to keep the source code as simple as possible. But automatic tests make the code more complex, in order to introduce clean coding boundaries. This extra cost cannot be avoided and it's higher in Go, comparing to VM-based languages, for example Python and Java.

Automatic tests are important for open source projects, because there aren't dedicated manual testers, who make manual tests for new functions and regression tests. The tradeoff between untested features (risk) and use case coverage (cost) is decision of the project (repo) maintainers.

When I watched the source code at the first time, I realized, the risk <--> cost decision was low code (use case) coverage, because the boundaries (Go interfaces) were missing for high code coverage. So I skipped the automatic test. 

But when @crobby asked me to write automatic tests, I did it. There were some new/changed codes, which can be tested by unit tests with low effort. But there were some new/changed codes which need too much effort to write unit tests, because the code should be restructured (boundaries with Go interfaces), before writing unit tests.

Instead of the costly restructuring, I introduced function tests, which emulate a Rancher server responses with httptest.Server. The `TestMasterClientNewManagementClient` is also a function test, not a simple unit test, which checks only if master client is not nil. It's a function test, which covers more codes, including communication with Rancher server. During running the test, the usage of cliclient.DefaultHTTPClient is also tested by mock (the test is failed, if it's not used).

The introduced function tests look superfluous at the first time, but the this investment will return. This kind of tests can be improved to component tests, which start a `cli.Command` and cover most of the code.

The decision is in the hands of the maintainers: turning to the function and component tests, or staying at only unit test. Please notice me the unified decision of the maintainers and I will change the mocks to `go.uber.org/mock` or I will remove all the function tests and its related code changes.

crobby:
Closing in favor of https://github.com/rancher/cli/pull/406.
@pgillich Thank you so much for bringing attention to this and your contribution.

PR #398, opened 2024-10-17
[2.11] Avoid permissions errors on Windows

HarrisonWAffel:
### Issue: https://github.com/rancher/rancher/issues/46769

This PR updates `GetFilePermissionWarnings` to ignore Windows platforms. The current file permission checks are not appropriate for Windows as Go does not handle permission bits in the same way as it does on Linux. For Windows, Go will only use the owner bit, effectively turning `600` into `666`. Access Control Lists are the intended means of securing files, and the cli currently sets the ACLs properly, only allowing access to the local system and administrators.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -6,6 +6,7 @@ import (
 	"net/url"
 	"os"
 	"path/filepath"
+	"runtime"
 	"strings"
 
 	"github.com/sirupsen/logrus"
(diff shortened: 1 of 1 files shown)


crobby (approved):
Sounds reasonable.  Thanks for the explanation.

jiaqiluo (approved):

PR #399, opened 2024-10-28
Improve cluster|project list-members subcommands

pmatseykanets:
Ref: https://github.com/rancher/rancher/issues/47692

### DIFF
--- a/cmd/cluster.go
+++ b/cmd/cluster.go
@@ -4,11 +4,13 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"io"
 	"slices"
 	"strconv"
 	"strings"
 
 	"github.com/rancher/cli/cliclient"
+	"github.com/rancher/norman/types"
 	managementClient "github.com/rancher/rancher/pkg/client/generated/management/v3"
 	"github.com/sirupsen/logrus"
 	"github.com/urfave/cli"
@@ -182,9 +184,22 @@ func ClusterCommand() cli.Command {
 				Action: listClusterRoles,
 			},
 			{
-				Name:   "list-members",
-				Usage:  "List current members of the cluster",
-				Action: listClusterMembers,
+				Name:  "list-members",
+				Usage: "List current members of the cluster",
+				Action: func(cctx *cli.Context) error {
+					client, err := GetClient(cctx)
+					if err != nil {
+						return err
+					}
+
+					return listClusterMembers(
+						cctx,
+						cctx.App.Writer,
+						client.UserConfig,
+						client.ManagementClient.ClusterRoleTemplateBinding,
+						client.ManagementClient.Principal,
+					)
+				},
 				Flags: []cli.Flag{
 					cli.StringFlag{
 						Name:  "cluster-id",
@@ -587,33 +602,30 @@ func listClusterRoles(ctx *cli.Context) error {
 	return listRoles(ctx, "cluster")
 }
 
-func listClusterMembers(ctx *cli.Context) error {
-	c, err := GetClient(ctx)
-	if err != nil {
-		return err
-	}
+type crtbLister interface {
+	List(opts *types.ListOpts) (*managementClient.ClusterRoleTemplateBindingCollection, error)
+}
 
-	clusterID := c.UserConfig.FocusedCluster()
+type userConfig interface {
+	FocusedCluster() string
+	FocusedProject() string
+}
+
+func listClusterMembers(ctx *cli.Context, out io.Writer, config userConfig, crtbs crtbLister, principals principalGetter) error {
+	clusterID := config.FocusedCluster()
 	if ctx.String("cluster-id") != "" {
 		clusterID = ctx.String("cluster-id")
 	}
 
 	filter := defaultListOpts(ctx)
 	filter.Filters["clusterId"] = clusterID
-	bindings, err := c.ManagementClient.ClusterRoleTemplateBinding.List(filter)
+
+	bindings, err := crtbs.List(filter)
 	if err != nil {
 		return err
 	}
 
-	userFilter := defaultListOpts(ctx)
-	users, err := c.ManagementClient.User.List(userFilter)
-	if err != nil {
-		return err
-	}
-
-	userMap := usersToNameMapping(users.Data)
-
-	var b []RoleTemplateBinding
+	rtbs := make([]RoleTemplateBinding, 0, len(bindings.Data))
 
 	for _, binding := range bindings.Data {
 		parsedTime, err := createdTimetoHuman(binding.Created)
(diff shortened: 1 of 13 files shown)


cmd/common.go
```diff
@@ -121,27 +121,60 @@ func listRoles(ctx *cli.Context, context string) error {
...
 }
 
+type principalGetter interface {
+	ByID(id string) (*managementClient.Principal, error)
+}
+
+func getMemberNameFromPrincipal(principals principalGetter, principalID string) string {
+	principal, err := principals.ByID(url.PathEscape(principalID))
+	if err != nil {
+		principal = parsePrincipalID(principalID)
```
bigkevmcd, line 147: Should this maybe only happen if it's not found? Feels a little odd to show the user details if we got a network error communicating with the cluster?

cmd/common_test.go
```diff
@@ -1,71 +1,201 @@
...
-			map[string]interface{}{"fooBar": "hello"},
-			map[string]interface{}{"fooBar": "hello"},
+			map[string]any{"foo_bar": "hello", "backup_config": map[string]any{"hello_world": true}, "config_id": 123},
+			map[string]any{"fooBar": "hello", "backupConfig": map[string]any{"helloWorld": true}, "configId": 123},
 		},
+	}
+
+	for _, test := range tests {
+		test := test
+		t.Run("", func(t *testing.T) {
```
bigkevmcd, line 99: Maybe instead of "" here, we could output the test number?

```go
for i, test := range tests {
    test := test
    t.Run(fmt.Sprintf("TestConvertSnakeCaseKeysToCamelCase-%d", i), func(...) { 
    })
}
```

cmd/common_test.go
```diff
@@ -1,71 +1,201 @@
...
+			},
 		},
 	}
 
-	for _, tc := range cases {
-		convertSnakeCaseKeysToCamelCase(tc.input)
-		c.Assert(tc.input, check.DeepEquals, tc.renamed)
+	for _, test := range tests {
+		test := test
+		t.Run("", func(t *testing.T) {
```
bigkevmcd, line 137: How about outputting `test.id` instead of `""` ?

cmd/project_test.go
```diff
@@ -0,0 +1,99 @@
...
+	}
+
+	flagSet := flag.NewFlagSet("test", flag.ContinueOnError)
+	cctx := cli.NewContext(nil, flagSet, nil)
+
+	var out bytes.Buffer
+
+	err := listProjectMembers(cctx, &out, userConfig, prtbs, principals)
+	require.NoError(t, err)
+	require.NotEmpty(t, out)
```
bigkevmcd, line 88: nit: you don't need `require.NotEmpty(t, out)` as you have a deeper assertion below.

bigkevmcd (changes requested):
Looks good to me, some minor nits and one thing that I think should be fixed.

alegrey91 (approved):
LGTM

cmd/project.go
```diff
@@ -292,49 +312,52 @@ func listProjectRoles(ctx *cli.Context) error {
...
-	c, err := GetClient(ctx)
-	if err != nil {
-		return err
-	}
+type prtbLister interface {
+	List(opts *types.ListOpts) (*managementClient.ProjectRoleTemplateBindingCollection, error)
+}
 
-	projectID := c.UserConfig.Project
+func listProjectMembers(ctx *cli.Context, out io.Writer, config userConfig, prtbs prtbLister, principals principalGetter) error {
```
enrichman, line 319: I love the change of the signature to make tests easier (and possible). I would try to avoid passing around the `cli.Context` if not really needed, and pass the needed arguments instead. I.e. something like a more generic `ListProjectMembersConfig` with the possible configuration. Not sure if we need a userConfig, or we can just put everything in one.

```go
type ListProjectMembersConfig struct {
	ProjectID	string
	Format		bool
	Quiet		bool
}

listProjectMembers(config ListProjectMembersConfig, out io.Writer, prtbs prtbLister, principals principalGetter) {
	// do stuff
}
```

Maybe we can even put the whole `TableWriterConfig` in the Configc, and it will be cleaner:

```go
type ListProjectMembersConfig struct {
	ProjectID		string
	TableWriterConfig	*TableWriterConfig
}

listProjectMembers(config ListProjectMembersConfig, prtbs prtbLister, principals principalGetter) {
	// do stuff
}
```

We could also have a constructor, helper and validation to create the config from the cli.Context, if needed.

What do you think?

cmd/common.go
```diff
@@ -121,27 +121,60 @@ func listRoles(ctx *cli.Context, context string) error {
 	return writer.Err()
 }
 
-func listRoleTemplateBindings(ctx *cli.Context, b []RoleTemplateBinding) error {
-	writer := NewTableWriter([][]string{
+func listRoleTemplateBindings(ctx *cli.Context, writerConfig *TableWriterConfig, rtbs []RoleTemplateBinding) error {
```
enrichman, line 124: If I'm not wrong the `cli.Context` is not needed anymore.

enrichman (commented):
Love the changes, and tests. Thanks!

cmd/common.go
pmatseykanets: You're absolutely right.

cmd/project_test.go
pmatseykanets: This was intentional. There is no need to do any other work if the output is empty. This assertion is cheap.

cmd/common_test.go
pmatseykanets: Of course.

cmd/common_test.go
pmatseykanets: Yep, just the number will do. There is no need to pass the test name. Go subtests always carry the parent test's name in the output.

cmd/project.go
pmatseykanets: This is a good idea, but I would suggest we look holistically at all commands and see what the best way to refactor this. At the moment and for the purposes of this PR `cli.Context` is as good of a container as any struct to pass flags around. There is certain amount of common flags that are being handled like default list and format options that need to considered.

cmd/cluster.go
```diff
@@ -587,49 +607,57 @@ func listClusterRoles(ctx *cli.Context) error {
...
+	List(opts *types.ListOpts) (*managementClient.ClusterRoleTemplateBindingCollection, error)
+}
 
-	clusterID := c.UserConfig.FocusedCluster()
+type userConfig interface {
+	FocusedCluster() string
+	FocusedProject() string
+}
+
+func listClusterMembers(ctx *cli.Context, out io.Writer, config userConfig, crtbs crtbLister, principals principalGetter) error {
```
crobby, line 619: Great change for testability

crobby (approved):
lgtm, addresses the original issue and makes things a lot better overall...and adds tests.

cmd/project.go
enrichman: That's good to me. Let's try to not diverge too much. :)

enrichman (approved):

pmatseykanets:
Do not merge: this is slated for v2.11.

PR #406, opened 2024-12-05
Respect proxy settings and set http timeout in login and ssh commands

pmatseykanets:
Ref: https://github.com/rancher/rancher/issues/48321

Dedicated http clients in login and ssh commands that downloads certs and an ssh key respectively currently don't respect proxy settings and don't set http timeout.

This PR fixes that and makes use of Norman's API Client Timeout and ProxyURL options and makes it possible to specify those in the server config.

### DIFF
--- a/cmd/common_test.go
+++ b/cmd/common_test.go
@@ -1,11 +1,15 @@
 package cmd
 
 import (
+	"crypto/tls"
 	"fmt"
+	"net/http"
 	"net/url"
 	"strconv"
 	"testing"
+	"time"
 
+	"github.com/rancher/cli/config"
 	managementClient "github.com/rancher/rancher/pkg/client/generated/management/v3"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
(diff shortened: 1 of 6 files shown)


config/config.go
```diff
@@ -26,14 +26,16 @@ type Config struct {
...
+	AccessKey          string                     `json:"accessKey"`
+	SecretKey          string                     `json:"secretKey"`
+	TokenKey           string                     `json:"tokenKey"`
+	URL                string                     `json:"url"`
+	Project            string                     `json:"project"`
+	CACerts            string                     `json:"cacert"`
+	KubeCredentials    map[string]*ExecCredential `json:"kubeCredentials"`
+	KubeConfigs        map[string]*api.Config     `json:"kubeConfigs"`
+	ProxyURL           string                     `json:"proxyUrl"`
+	HTTPTimeoutSeconds int                        `json:"httpTimeoutSeconds"`
```
pmatseykanets, line 38: We probably don't need a finer precision for this. Thus `*Seconds`.

cmd/login.go
```diff
@@ -260,10 +261,25 @@ func getCertFromServer(ctx *cli.Context, cf *config.ServerConfig) (*cliclient.Ma
...
+	var proxy func(*http.Request) (*url.URL, error)
+	if cf.ProxyURL != "" {
+		proxyURL, err := url.Parse(cf.ProxyURL)
+		if err != nil {
+			return nil, fmt.Errorf("invalid proxy address %s: %w", cf.ProxyURL, err)
+		}
+		proxy = http.ProxyURL(proxyURL)
+	} else {
+		proxy = http.ProxyFromEnvironment
+	}
```
pmatseykanets, line 273: This can potentially be extracted into a helper to share between login and ssh commands.

cmd/login.go
```diff
@@ -260,10 +261,25 @@ func getCertFromServer(ctx *cli.Context, cf *config.ServerConfig) (*cliclient.Ma
...
+
 	tr := &http.Transport{
 		TLSClientConfig: &tls.Config{InsecureSkipVerify: true},
+		Proxy:           proxy,
 	}
 	client := &http.Client{Transport: tr}
+	if cf.HTTPTimeoutSeconds > 0 {
+		client.Timeout = time.Duration(cf.HTTPTimeoutSeconds) * time.Second
```
pmatseykanets, line 281: This can potentially be a method on ServerConfig.

alegrey91 (dismissed):

cmd/login.go
pmatseykanets: Done.

cmd/login.go
pmatseykanets: Done.

andreas-kupries (approved):

alegrey91 (dismissed):
LGTM

cmd/common_test.go
```diff
@@ -200,3 +204,91 @@ func TestGetMemberNameFromPrincipal(t *testing.T) {
 		})
 	}
 }
+
+func TestNewHTTPClient(t *testing.T) {
```
pmatseykanets, line 208: FYI Can use `t.Parallel()` because of `t.Setenv()`.

alegrey91 (approved):

crobby (approved):
lgtm.  Tests seem reasonable.

pmatseykanets:
> Should link the issue
We never had one, so I opened an new one.

pmatseykanets:
Start local proxy
```
mitmproxy -p 8088
```
Set proxy and run rancher cli
```
export http_proxy='localhost:8088'
export https_proxy='localhost:8088'

rancher cluster ls
CURRENT   ID             STATE     NAME         PROVIDER   NODES     CPU       RAM            PODS
*         c-m-tbgzfbgf   active    downstream   K3S        1         0.55/12   0.32/7.65 GB   12/110
          local          active    local        K3S        1         0.20/12   0.14/7.65 GB   11/110
```
image

andreas-kupries (approved):

PR #410, opened 2025-01-08
Respect proxy settings and timeout in token command

pmatseykanets:
Ref https://github.com/rancher/rancher/issues/48631

The dedicated http clients in the `token` command that are used for various needs currently don't respect proxy and timeout settings.

This PR makes sure that the http client is reused and respect proxy and timeout settings.

### DIFF
--- a/cmd/kubectl_token.go
+++ b/cmd/kubectl_token.go
@@ -392,20 +392,16 @@ func basicAuth(client *http.Client, input *LoginInput) (managementClient.Token, 
 	resp, respBody, err := doRequest(client, req)
 	if err == nil && resp.StatusCode != http.StatusOK {
 		err = fmt.Errorf("unexpected http status code %d", resp.StatusCode)
-	}
-	if err != nil {
-		return token, nil
-	}
 
-	apiError := map[string]interface{}{}
-	err = json.Unmarshal(respBody, &apiError)
+		apiError := map[string]interface{}{}
+		if rerr := json.Unmarshal(respBody, &apiError); rerr == nil {
+			if responseType := apiError["type"]; responseType == "error" {
+				err = fmt.Errorf("error logging in: code: [%v] message:[%v]", apiError["code"], apiError["message"])
+			}
+		}
+	}
 	if err != nil {
 		return token, err
-	}
-
-	if responseType := apiError["type"]; responseType == "error" {
-		return token, fmt.Errorf("error logging in: code: "+
-			"[%v] message:[%v]", apiError["code"], apiError["message"])
 	}
 
 	err = json.Unmarshal(respBody, &token)
--- a/cmd/kubectl_token.go
+++ b/cmd/kubectl_token.go
@@ -14,7 +14,7 @@ import (
 	"io"
 	"math/big"
 	"net/http"
-	url2 "net/url"
+	"net/url"
 	"os"
 	"os/signal"
 	"runtime"
@@ -135,28 +135,32 @@ func CredentialCommand() cli.Command {
 }
 
 func runCredential(ctx *cli.Context) error {
-	if ctx.Bool("delete") {
-		return deleteCachedCredential(ctx)
-	}
 	server := ctx.String("server")
 	if server == "" {
 		return errors.New("name of rancher server is required")
 	}
-	url, err := url2.Parse(server)
+
+	serverURL, err := url.Parse(server)
 	if err != nil {
 		return err
 	}
-	if url.Scheme == "" {
+	if serverURL.Scheme == "" {
 		server = fmt.Sprintf("https://%s", server)
 	}
+
 	userID := ctx.String("user")
 	if userID == "" {
 		return errors.New("user-id is required")
 	}
 	clusterID := ctx.String("cluster")
 
+	serverConfig, err := lookupServerConfig(ctx)
+	if err != nil {
+		return fmt.Errorf("error looking up server config: %w", err)
+	}
+
 	cachedCredName := fmt.Sprintf("%s_%s", userID, clusterID)
-	cachedCred, err := loadCachedCredential(ctx, cachedCredName)
+	cachedCred, err := loadCachedCredential(ctx, serverConfig, cachedCredName)
 	if err != nil {
 		customPrint(fmt.Errorf("LoadToken: %v", err))
 	}
@@ -173,12 +177,22 @@ func runCredential(ctx *cli.Context) error {
 		skipVerify:   ctx.Bool("skip-verify"),
 	}
 
-	newCred, err := loginAndGenerateCred(input)
+	tlsConfig, err := getTLSConfig(input.skipVerify, input.caCerts)
 	if err != nil {
 		return err
 	}
 
-	if err := cacheCredential(ctx, newCred, fmt.Sprintf("%s_%s", userID, clusterID)); err != nil {
+	client, err := newHTTPClient(serverConfig, tlsConfig)
+	if err != nil {
+		return err
+	}
+
+	newCred, err := loginAndGenerateCred(client, input)
+	if err != nil {
+		return err
+	}
+
+	if err := cacheCredential(ctx, serverConfig, cachedCredName, newCred); err != nil {
 		customPrint(fmt.Errorf("CacheToken: %v", err))
 	}
 
@@ -214,20 +228,15 @@ func deleteCachedCredential(ctx *cli.Context) error {
 	for _, key := range ctx.Args() {
 		customPrint(fmt.Sprintf("removing [%s]", key))
 		for _, server := range cf.Servers {
-			server.KubeCredentials[key] = nil
+			delete(server.KubeCredentials, key)
 		}
 	}
 
 	return cf.Write()
 }
 
-func loadCachedCredential(ctx *cli.Context, key string) (*config.ExecCredential, error) {
-	sc, err := lookupServerConfig(ctx)
-	if err != nil {
-		return nil, err
-	}
-
-	cred := sc.KubeToken(key)
+func loadCachedCredential(ctx *cli.Context, serverConfig *config.ServerConfig, key string) (*config.ExecCredential, error) {
+	cred := serverConfig.KubeToken(key)
 	if cred == nil {
 		return cred, nil
 	}
@@ -275,7 +284,7 @@ func lookupServerConfig(ctx *cli.Context) (*config.ServerConfig, error) {
 	return sc, nil
 }
 
-func cacheCredential(ctx *cli.Context, cred *config.ExecCredential, id string) error {
+func cacheCredential(ctx *cli.Context, serverConfig *config.ServerConfig, key string, cred *config.ExecCredential) error {
 	// cache only if valid
 	if cred.Status.Token == "" {
 		return nil
@@ -291,28 +300,17 @@ func cacheCredential(ctx *cli.Context, cred *config.ExecCredential, id string) e
 		return err
 	}
 
-	sc, err := lookupServerConfig(ctx)
-	if err != nil {
-		return err
+	if serverConfig.KubeCredentials == nil {
+		serverConfig.KubeCredentials = make(map[string]*config.ExecCredential)
 	}
-
-	if sc.KubeCredentials == nil {
-		sc.KubeCredentials = make(map[string]*config.ExecCredential)
-	}
-	sc.KubeCredentials[id] = cred
-	cf.Servers[server] = sc
+	serverConfig.KubeCredentials[key] = cred
+	cf.Servers[server] = serverConfig
 
 	return cf.Write()
 }
 
-func loginAndGenerateCred(input *LoginInput) (*config.ExecCredential, error) {
-	// setup a client with the provided TLS configuration
-	client, err := getClient(input.skipVerify, input.caCerts)
-	if err != nil {
-		return nil, err
-	}
-
-	authProviders, err := getAuthProviders(input.server)
+func loginAndGenerateCred(client *http.Client, input *LoginInput) (*config.ExecCredential, error) {
+	authProviders, err := getAuthProviders(client, input.server)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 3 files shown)


cmd/kubectl_token.go
```diff
@@ -135,28 +135,32 @@ func CredentialCommand() cli.Command {
...
 
 func runCredential(ctx *cli.Context) error {
-	if ctx.Bool("delete") {
-		return deleteCachedCredential(ctx)
-	}
 	server := ctx.String("server")
 	if server == "" {
 		return errors.New("name of rancher server is required")
 	}
-	url, err := url2.Parse(server)
```
pmatseykanets, line 145: Avoid shadowing the `url` package.

cmd/kubectl_token.go
```diff
@@ -173,12 +177,22 @@ func runCredential(ctx *cli.Context) error {
...
+	tlsConfig, err := getTLSConfig(input.skipVerify, input.caCerts)
 	if err != nil {
 		return err
 	}
 
-	if err := cacheCredential(ctx, newCred, fmt.Sprintf("%s_%s", userID, clusterID)); err != nil {
+	client, err := newHTTPClient(serverConfig, tlsConfig)
+	if err != nil {
+		return err
+	}
```
pmatseykanets, line 188: Reuse the common http client for all subsequent requests.

cmd/kubectl_token.go
```diff
@@ -173,12 +177,22 @@ func runCredential(ctx *cli.Context) error {
...
 	}
 
-	newCred, err := loginAndGenerateCred(input)
+	tlsConfig, err := getTLSConfig(input.skipVerify, input.caCerts)
 	if err != nil {
 		return err
 	}
 
-	if err := cacheCredential(ctx, newCred, fmt.Sprintf("%s_%s", userID, clusterID)); err != nil {
+	client, err := newHTTPClient(serverConfig, tlsConfig)
```
pmatseykanets, line 185: Reuse the common http client for all subsequent requests.

cmd/kubectl_token.go
```diff
@@ -214,20 +228,15 @@ func deleteCachedCredential(ctx *cli.Context) error {
 	for _, key := range ctx.Args() {
 		customPrint(fmt.Sprintf("removing [%s]", key))
 		for _, server := range cf.Servers {
-			server.KubeCredentials[key] = nil
```
pmatseykanets, line 217: This was prone to panics.

cmd/kubectl_token.go
```diff
@@ -402,15 +408,17 @@ func basicAuth(input *LoginInput) (managementClient.Token, error) {
...
+	err = json.Unmarshal(respBody, &token)
 	if err != nil {
 		return token, err
 	}
+
 	return token, nil
 }
 
-func samlAuth(input *LoginInput, client *http.Client) (managementClient.Token, error) {
+func samlAuth(client *http.Client, input *LoginInput) (managementClient.Token, error) {
```
pmatseykanets, line 419: The idiomatic approach is to list arguments in the order of increasing specificity. This way all dependencies come first.

cmd/kubectl_token.go
```diff
@@ -489,55 +497,54 @@ func samlAuth(input *LoginInput, client *http.Client) (managementClient.Token, e
...
-				customPrint(fmt.Errorf("DeleteToken: %v", err))
+				// Log the error and move on.
+				customPrint(fmt.Errorf("error deleting auth token: %s", err))
 			}
-			defer res.Body.Close()
 
 			return token, nil
 
 		case <-timeout.C:
-			break
```
pmatseykanets, line 512: This was no-op as it doesn't break out of the loop.

cmd/kubectl_token.go
```diff
@@ -489,55 +497,54 @@ func samlAuth(input *LoginInput, client *http.Client) (managementClient.Token, e
...
+			customPrint("timed out waiting for the auth token")
+			break loop
 
 		case <-interrupt:
-			customPrint("received interrupt")
-			break
```
pmatseykanets, line 516: Same here.

cmd/kubectl_token.go
```diff
@@ -671,30 +662,19 @@ func getTLSConfig(skipVerify bool, caCerts string) (*tls.Config, error) {
 	return config, nil
 }
 
-func request(method, url string, body io.Reader) ([]byte, error) {
-	var response []byte
-
-	req, err := http.NewRequest(method, url, body)
+func doRequest(client *http.Client, req *http.Request) (*http.Response, []byte, error) {
```
pmatseykanets, line 665: Return the `http.Response` as the consuming code should be checking the returned status code in most/all cases.

cmd/kubectl_token.go
```diff
@@ -135,28 +135,32 @@ func CredentialCommand() cli.Command {
 }
 
 func runCredential(ctx *cli.Context) error {
-	if ctx.Bool("delete") {
-		return deleteCachedCredential(ctx)
-	}
```
pmatseykanets, line 140: Seems like it might've been a flag in the past, which is no longer true. We use the `delete` subcommand for this.

cmd/kubectl_token.go
```diff
@@ -434,45 +442,45 @@ func samlAuth(input *LoginInput, client *http.Client) (managementClient.Token, e
...
 	if err != nil {
 		return token, err
 	}
+	getReq.Header.Set("content-type", "application/json")
+	getReq.Header.Set("accept", "application/json")
 
-	req.Header.Set("content-type", "application/json")
-	req.Header.Set("accept", "application/json")
-
-	client.Timeout = 300 * time.Second
```
pmatseykanets, line 445: This seems unnecessary. There is nothing specific or unusual about requesting an auth token that might require a custom 5 min timeout.

cmd/kubectl_token.go
```diff
@@ -489,55 +497,54 @@ func samlAuth(input *LoginInput, client *http.Client) (managementClient.Token, e
...
-				return token, err
+			// Delete the auth token.
+			resp, _, err := doRequest(client, deleteReq)
+			if err == nil && resp.StatusCode != http.StatusOK && resp.StatusCode != http.StatusNotFound {
+				err = fmt.Errorf("unexpected http status code %d", resp.StatusCode)
 			}
-			req.Header.Set("content-type", "application/json")
-			req.Header.Set("accept", "application/json")
-
-			client.Timeout = 150 * time.Second
```
pmatseykanets, line 500: Same here. There is nothing specific or unusual about this request that might require a custom timeout.

cmd/kubectl_token.go
enrichman: One of the reason because we should try to avoid passing the `*cli.Context`, and have a `Config` struct to pass. But that's a refactor for another day.

cmd/kubectl_token.go
enrichman: Nice!

cmd/kubectl_token.go
```diff
@@ -381,18 +379,26 @@ func basicAuth(input *LoginInput) (managementClient.Token, error) {
...
-	response, err := request(http.MethodPost, url, bytes.NewBufferString(body))
+	req, err := http.NewRequest(http.MethodPost, loginURL, bytes.NewBufferString(reqBody))
+	if err != nil {
+		return token, fmt.Errorf("error creating request: %w", err)
+	}
+
+	resp, respBody, err := doRequest(client, req)
+	if err == nil && resp.StatusCode != http.StatusOK {
+		err = fmt.Errorf("unexpected http status code %d", resp.StatusCode)
+	}
```
enrichman, line 395: I feel having the err check before is more idiomatic, wdyt?

```go
resp, respBody, err := doRequest(client, req)
if err != nil {
	return token, nil
}
if resp.StatusCode != http.StatusOK {
	return token, fmt.Errorf("unexpected http status code %d", resp.StatusCode)
}
```

I'm also confused about the `return token, nil`. Is it ok? Should we at least log the error?

cmd/kubectl_token.go
```diff
@@ -434,45 +442,45 @@ func samlAuth(input *LoginInput, client *http.Client) (managementClient.Token, e
...
-	customPrint(fmt.Sprintf("\nLogin to Rancher Server at %s \n", loginRequest))
+	customPrint(fmt.Sprintf("\nLogin to Rancher Server at %s \n", loginURL))
 
 	interrupt := make(chan os.Signal, 1)
 	signal.Notify(interrupt, os.Interrupt)
 
-	// timeout for user to login and get token
-	timeout := time.NewTicker(15 * time.Minute)
+	// Timeout for the login flow.
+	timeout := time.NewTimer(15 * time.Minute)
```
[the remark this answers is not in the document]
enrichman: I feel there is a better way to do this poll and timeout, but I wouldn't try to change the logic without tests here.

cmd/kubectl_token_oauth.go
```diff
@@ -84,15 +85,24 @@ func rancherLogin(input *LoginInput, provider TypedProvider, oauthToken *oauth2.
...
-	b, err := request(http.MethodPost, url, bytes.NewBuffer(jsonBody))
+	req, err := http.NewRequest(http.MethodPost, url, bytes.NewBuffer(jsonBody))
+	if err != nil {
+		return nil, fmt.Errorf("error creating request: %w", err)
+	}
+
+	resp, respBody, err := doRequest(client, req)
+	if err == nil && resp.StatusCode != http.StatusOK {
+		return nil, fmt.Errorf("unexpected http status code %d", resp.StatusCode)
+	}
```
enrichman, line 96: Same as above, can we invert the check?

enrichman (commented):
A couple of comments, but it LGTM

cmd/kubectl_token.go
pmatseykanets: IMHO passing the context or not is a different story.

The issue here is that it seems like that delete was a flag 
`rancher token --delete <name>`
and then turned into an argument (subcommand)
`rancher token delete <name>`

cmd/kubectl_token.go
pmatseykanets: The only idiomatic part is to explicitly check for the error if one is returned to the caller, which we do. There is no prescribed way of doing it though.
In this particular case if the request succeeds `err == nil` but the http status code is anything but the OK we want to fail and so we assign an error first L393-L395.
Then we cover both cases and check for `err != nil`.

cmd/kubectl_token_oauth.go
pmatseykanets: This was a bad copy-paste from one of the my previous iterations. We shouldn't invert anything. All we need is to set the err. Fixed in 649c29df36cec21939cc05ce66114d8e4d11cd97.

cmd/kubectl_token.go
```diff
@@ -381,18 +379,26 @@ func basicAuth(input *LoginInput) (managementClient.Token, error) {
...
+	if err != nil {
+		return token, fmt.Errorf("error creating request: %w", err)
+	}
+
+	resp, respBody, err := doRequest(client, req)
+	if err == nil && resp.StatusCode != http.StatusOK {
+		err = fmt.Errorf("unexpected http status code %d", resp.StatusCode)
+	}
 	if err != nil {
 		return token, nil
```
andreas-kupries, line 397: I do not understand this here. We have an error here (see 2 lines above), and return the token, but not the error.
I would have expected
```suggestion
		return nil, err
```
here.

cmd/kubectl_token.go
```diff
@@ -381,18 +379,26 @@ func basicAuth(input *LoginInput) (managementClient.Token, error) {
...
 	}
 
 	apiError := map[string]interface{}{}
-	err = json.Unmarshal(response, &apiError)
+	err = json.Unmarshal(respBody, &apiError)
 	if err != nil {
 		return token, err
```
andreas-kupries, line 403: Is it ok to return both error and token here ? Should it be the error alone instead, without the token ?

cmd/kubectl_token.go
```diff
@@ -402,15 +408,17 @@ func basicAuth(input *LoginInput) (managementClient.Token, error) {
 			"[%v] message:[%v]", apiError["code"], apiError["message"])
 	}
 
-	err = json.Unmarshal(response, &token)
+	err = json.Unmarshal(respBody, &token)
 	if err != nil {
 		return token, err
```
andreas-kupries, line 413: Ditto: Is it ok to return both error and token here ? Should it be the error alone instead, without the token ?

cmd/kubectl_token.go
```diff
@@ -434,45 +442,45 @@ func samlAuth(input *LoginInput, client *http.Client) (managementClient.Token, e
...
 
+loop:
 	for {
 		select {
 		case <-poll.C:
-			res, err := client.Do(req)
+			// Fetch the auth token.
+			_, respBody, err := doRequest(client, getReq)
 			if err != nil {
 				return token, err
```
andreas-kupries, line 481: Ditto: Is it ok to return both error and token here ? Should it be the error alone instead, without the token ?

cmd/kubectl_token.go
```diff
@@ -434,45 +442,45 @@ func samlAuth(input *LoginInput, client *http.Client) (managementClient.Token, e
...
-			content, err := io.ReadAll(res.Body)
-			if err != nil {
-				res.Body.Close()
-				return token, err
-			}
-			res.Body.Close()
-			err = json.Unmarshal(content, &token)
+			err = json.Unmarshal(respBody, &token)
 			if err != nil {
 				return token, err
```
andreas-kupries, line 485: Ditto: Is it ok to return both error and token here ? Should it be the error alone instead, without the token ?

andreas-kupries (changes requested):
Some questions.

cmd/kubectl_token.go
pmatseykanets: Good catch, this is a left-over from the previous error  handling logic.
Changed in 3338ec579d191906c6765c1ab0eada15bb21d7a5. We now attempt to parse the the error response only if the request comes back with the status is not OK. The apiError unmarshalling error is ignored deliberately.

cmd/kubectl_token.go
pmatseykanets: TL;DR Yes, it's. OK.
The `token` is of the value `managementClient.Token` not the reference type `*managementClient.Token` in this case. The returned value is ignored at the call site L338 in the case of an error as it should be.

cmd/kubectl_token.go
pmatseykanets: See above.

cmd/kubectl_token.go
pmatseykanets: See above.

andreas-kupries (approved):

enrichman (approved):

PR #412, opened 2025-02-11
Fix expected http status codes for different auth methods in token command

pmatseykanets:
Ref: https://github.com/rancher/rancher/issues/48631

Fix expected HTTP status codes from rancher endpoints responsible for logging in using different auth providers (basic auth, oauth, saml).


JonCrowther (approved):
Seems ok to me. I appreciate the improved error messages too!

crobby (dismissed):
Change seems reasonable.
Is there an issue for this that could be referenced?

crobby (approved):

PR #447, opened 2025-07-08
feat: support platform linux/arm64

salzig:
Using Docker (colima) on modern Apple Hardware runs container on linux/arm64. 

When did I stumble upon this? 
We use Container Images to "wrap" all necessary tools to interact with our rancher clusters. To build and install those images we use `mise`  (https://mise.jdx.dev) which resolves the correct dependency to download for us. Which is platform linux/arm64 that results in a `mise ERROR unsupported env: linux/arm64` cause there is no artefact for that platform provided.

### DIFF
--- a/scripts/build
+++ b/scripts/build
@@ -7,7 +7,7 @@ cd $(dirname $0)/..
 declare -A OS_ARCH_ARG
 
 OS_PLATFORM_ARG=(linux windows darwin)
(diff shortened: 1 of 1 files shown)


salzig:
To be honest, for us this is a Bug as it holds everyone in our team from building/using that image with the correct platform. As a workaround we can use a x86 based qemu image, which is slower.

pmatseykanets:
Seems reasonable.

PR #539, opened 2026-01-23
Add wrapper for helm commands

techknowlogick:
Note: I purposely did 1:1 copy of the `kubectl` wrapper rather than refactoring out the commands into a common function. I figured that I would start with this and possibly a discussion would arise on the preferred approach of this repo. I'm more than happy to refactor this into a common shared helper.

Especially with the above, as perhaps with a common shared helper, then maybe even a generic wrapper command for arbitrary wrapped commands. This is a possible future state, and I'd rather start discussions here.

### DIFF
--- a/cmd/helm.go
+++ b/cmd/helm.go
@@ -1,0 +1,119 @@
+package cmd
+
+import (
+	"fmt"
+	"os"
+	"os/exec"
+
+	"github.com/urfave/cli"
+	"k8s.io/client-go/tools/clientcmd"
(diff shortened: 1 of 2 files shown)


fis-max:
Are we not doing this?

When working with agents on clusters, the Rancher CLI is a very nice tool to have.
But even just having a single entry point to manage my Kubernetes contexts with one authorization is very convenient.

@samjustus, why are you planning to kill it?

PR #591, opened 2026-04-24
Fixes and Tweak for AzureAD OAuth

binaryoverload:
- On Windows, the authorization URL was being split by & and failing - changed to use a method which works with URLs with &
- ~~The default port was set to be random, which I think should be instead set to a static port so the default configuration has a static callback URL which can be configured~~ Azure AD works with random ports, this change has been reverted

### DIFF
--- a/cmd/kubectl_token_oauth.go
+++ b/cmd/kubectl_token_oauth.go
@@ -52,7 +52,7 @@ func oauthAuth(client *http.Client, input *LoginInput, provider TypedProvider, u
 func getCallbackPort() (int, error) {
 	env := os.Getenv("CATTLE_OAUTH_CALLBACK_PORT")
 	if env == "" {
-		return 0, nil // Use random port
+		return 8888, nil // Default port
 	}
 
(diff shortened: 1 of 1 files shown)


cmd/kubectl_token_oauth.go
```diff
@@ -52,7 +52,7 @@ func oauthAuth(client *http.Client, input *LoginInput, provider TypedProvider, u
 func getCallbackPort() (int, error) {
 	env := os.Getenv("CATTLE_OAUTH_CALLBACK_PORT")
 	if env == "" {
-		return 0, nil // Use random port
+		return 8888, nil // Default port
```
andreas-kupries: I am worried that the fixed port is a security issue using a random port prevents.
🔔 @macedogm

binaryoverload: I've just tested this with Azure - and the ephemeral port does actually work with just `http://localhost` as the redirect URL

I will revert this change

andreas-kupries (approved):

PR #657, opened 2026-08-26
Open the OAuth authorize URL on Windows without a shell

pmatseykanets:
## Problem

On Windows, openBrowser launched the browser via `cmd /c start "" <url>`. cmd.exe parses `&` as a command separator, so authorize URLs with multiple query parameters (e.g. Azure AD's client_id, scope, response_type) were truncated before reaching the browser, causing AADSTS900144 failures.

## Solution

`openBrowser` now delegates OS-specific launching to a new `osOpenURL` function. The Windows implementation uses the `ShellExecute` API, passing the URL as an opaque argument with no command line built, so it can't be reinterpreted by a shell. macOS and Linux keep using `exec.Command` with the URL as a single `argv` element, unchanged in behavior.

### DIFF
--- a/cmd/kubectl_token_oauth_windows_test.go
+++ b/cmd/kubectl_token_oauth_windows_test.go
@@ -1,0 +1,43 @@
+//go:build windows
+
+package cmd
+
+import (
+	"testing"
+
+	"golang.org/x/sys/windows"
+)
+
+// TestOSOpenURLDoesNotShellInterpret guards the fix for AADSTS900144: it stubs
+// the ShellExecute call and asserts osOpenURL hands the URL to the API
+// opaquely — the full URL, metacharacters and all, as the target, with no
+// command-line arguments — so there is no shell for a server-supplied
+// authorize URL to be truncated by. The old `cmd /c start "" <url>` path built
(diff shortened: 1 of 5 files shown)


bigkevmcd:
It looks like we could do with an `exec.Command` call?

https://github.com/Azure/unbounded/blob/7056e0651c0924b8bedd5104468dff63f42bcc9f/cmd/kubectl-unbounded/app/net/proxy.go#L345-L358

But I can find other projects that use the same technique... https://github.com/pkg/browser/blob/master/browser_windows.go

How did you choose which to use?

andreas-kupries (approved):

bigkevmcd (approved):

pmatseykanets:
> It looks like we could do with an exec.Command call?

We can, but `ShellExecute` seems to be the recommended approach that avoids spawning a subprocess.

cmd/kubectl_token_oauth_windows_test.go
```diff
@@ -0,0 +1,43 @@
+//go:build windows
```
pmatseykanets, line 1: FYI: To actually utilize this we'll need a Windows runner.
We'll consider this separately.

